PR #26, opened 2014-02-21
Experimental PR that will test comtypes on travis-ci under wine

cfarrow:


junkmd:
Somehow this branch was still remained, so I deleted it.

In case you still need it, you can use the `Restore branch` button on GitHub.

PR #75, opened 2015-01-15
Fixes TypeError when wrapping `esriSystem.olb`

maphew:
More details at http://gis.stackexchange.com/questions/37672/arcobjects-comtypes-at-10-1. Comtypes is the most popular route to using the Esri ArcObjects com library for functions that are not exposed in the official ArcPy library.

### DIFF
--- a/comtypes/automation.py
+++ b/comtypes/automation.py
@@ -858,6 +858,9 @@ _ctype_to_vartype = {
     # such an array cannot be created.
     POINTER(VARIANT): VT_BYREF|VT_VARIANT,
 
+    POINTER(BSTR): VT_BYREF|VT_BSTR,
+        #fix for Esri ArcObjects (esriSystem.olb)
+
     # These are not yet implemented:
 ##    POINTER(IUnknown): VT_UNKNOWN,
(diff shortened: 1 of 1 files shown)


cfarrow:
Thank you for the patch. I am looking for a way to write a test for this before merging the change into the codebase. I always prefer writing tests against freely available libraries so there is little overhead to running the tests. I'll look at alternative options if this turns out to be infeasible.

maphew:
Thanks for considering it. I'm happy to run tests you or anyone else could provide against my environment and share the results. Unfortunately I'm near clueless how to write proper tests for something like this myself.

cfarrow:
Sorry for the long delay. I've determined that this will do good without harm. I have been able to write a simple test, so I'm merging it. Thanks!

maphew:
Oh this is wonderful, thank you!

Matt

cfarrow:
I meant to say _unable_ to write a simple test. Nonetheless, I understand the patch and it will break anything. Thanks again.

markkness:
And I think you mean it will NOT break anything :)

cfarrow:
This is what happens when I try to do important things on the weekend. :D

PR #78, opened 2015-05-07
check for IOError in _is_writeable

RobinD42:
IOError is raised on windows there there is no write permission in the requested dir.


cfarrow:
LGTM. Thanks!

PR #95, opened 2016-02-19
Fix for Issue 92 - generated code has Python 3 syntax error

laripk:
This only fixes the narrow case of EnumValues potentially being python keywords.

I've only tested it against the indicated problem child, MSHTML (for those of us using pywebview on python3), which only hits the keywords `True` and `False`.

I included the debug statement so those using other COM libraries will see what gets changed and be forewarned (just in case the name change messes up the COM interaction).

Many thanks to @takluyver for his detailed Issue report.

### DIFF
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -2,8 +2,8 @@ import types
 import sys
 import os
 
-# comtypes version numbers follow semver (http://semver.org/)
-__version__ = "1.1.3-0.fixIssue92"
+# comtypes version numbers follow semver (http://semver.org/) and PEP 440
+__version__ = "1.1.3.dev0"
 
 import logging
 class NullHandler(logging.Handler):
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -3,7 +3,7 @@ import sys
 import os
 
 # comtypes version numbers follow semver (http://semver.org/)
-__version__ = "1.1.2"
+__version__ = "1.1.3-0.fixIssue92"
 
 import logging
 class NullHandler(logging.Handler):
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -9,6 +9,8 @@ import comtypes.client._generate
 
 version = "$Rev$"[6:-2]
 
+__warn_on_munge__ = __debug__
+
 
 class lcid(object):
     def __repr__(self):
@@ -348,7 +350,7 @@ class Generator(object):
         value = int(tp.value)
         if keyword.iskeyword(tp.name):
             # XXX use logging!
-            if __debug__:
+            if __warn_on_munge__:
                 print "# Fixing keyword as EnumValue for %s" % tp.name
             tp.name += "_"
         print >> self.stream, \
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -2,6 +2,7 @@
 # libraries.
 import os
 import cStringIO
+import keyword
 from comtypes.tools import typedesc
 import comtypes.client
 import comtypes.client._generate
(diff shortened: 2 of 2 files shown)


comtypes/tools/codegenerator.py
```diff
@@ -345,6 +346,11 @@ def ArrayType(self, tp):
     _enumvalues = 0
     def EnumValue(self, tp):
         value = int(tp.value)
+        if keyword.iskeyword(tp.name):
+            # XXX use logging!
+            if __debug__:
```
takluyver: I wouldn't rely on the `__debug__` flag here - [from the docs](https://docs.python.org/3/library/constants.html#__debug__) it's True unless Python was run with `-O`. In practice, no-one I know runs Python with `-O`, so it will ~always print the message.

takluyver:
Thanks for working out a fix for this, @laripk !

comtypes/__init__.py
```diff
@@ -3,7 +3,7 @@
 import os
 
 # comtypes version numbers follow semver (http://semver.org/)
-__version__ = "1.1.2"
+__version__ = "1.1.3-0.fixIssue92"
```
cfarrow: It's not noted here, but I think it is best to follow PEP 440. I recommend `1.1.3.dev0`.

comtypes/tools/codegenerator.py
cfarrow: I agree.

comtypes/__init__.py
laripk: I'm new to Python, and I missed that PEP. 
I've updated it.
Thanks.

cfarrow: Thank you.

comtypes/tools/codegenerator.py
laripk: While I'm not sure that having it always print the message would be bad in this particular case (because I strongly suspect that any COM object trying to overwrite a Python keyword that isn't `True` or `False` will cause problems), I agree that that wasn't my intention.

Since I haven't figured out how to use `logging` yet, I was following a pattern I saw in another file, however I skipped over the `__verbose__ = __debug__` step at the top of the file.

What would you recommend I do here?

takluyver: It doesn't look like there's any logging already in this file, so I'd probably just leave it out. Or you could put the print statement in but commented out, so someone could uncomment it to debug.

laripk: No, there wasn't any logging in this file but there is in other parts of the library.

I don't like the idea of simply commenting out the print statement, because if/when someone does have a COM object that uses a Python keyword other than `True` or `False` they are not going to know to look for this.

The pattern I was following was the use of the `__verbose__` flag in `client._generate.py`. I like this precedent because it only flags and prints the code generation. Note that `client._generate.py` also  uses `logging` but doesn't for those messages.

Perhaps if I followed that pattern properly (adding the flag at the top), but used a name that better expresses my intentions...

laripk: How does [that](https://github.com/laripk/comtypes/commit/81f65d9bf98449345100404e944b2ef835a8665e) look?

takluyver: I don't personally like the `__verbose__ = __debug__` pattern, because it has exactly the same problem I mentioned before - in practice, almost everyone will have verbose output enabled. But it's not my library, so if that's what they're doing...

cfarrow: I'd rather consistently use logging, but that should be handled in a separate issue. I think sticking with an established pattern is fine for now.

laripk: created issue #96 for you

cfarrow:
Indeed. Thank you.

Issue #63, opened 2014-06-11
`VARIANT._get_value` cannot unpack arrays of arbitrary type


cfarrow:
The `VARIANT._get_value` method does not unpack arrays of arbitrary type. As seen [here](https://github.com/enthought/comtypes/blob/f269c84159372799cd0f8d7b0ee23f547b0b21ec/comtypes/automation.py#L455), it is assumed that the unpacked type is in the limited `_vartype_to_ctype` dictionary. This is likely the cause behind reclosedev/pyautocad#6. The code should be expanded to handle more types.

PR #104, opened 2016-04-09
Addresses #63 by unpacking VT_ARRAYs of VT_RECORDs

mementum:
1. If regular array unpacking fails (item not present in vartype_to_ctype, it is checked if the array is a an array of records

2. The record information is extracted (with the new SafeArrayGetRecordInfo) and used to cast the array to the proper type (from the containing tlib)

3. Unpacking needs to copy the bytes, because if the record contains items like BSTR, the addresses of the 1st BSTRs will be overwritten during the generation of the subsequent items). See: https://codebright.wordpress.com/2010/06/20/ctypes-and-addressof/

### DIFF
--- a/comtypes/_safearray.py
+++ b/comtypes/_safearray.py
@@ -69,6 +69,18 @@ def SafeArrayGetVartype(pa):
     _SafeArrayGetVartype(pa, result)
     return result.value
 
+def SafeArrayGetRecordInfo(pa):
+    # Defined here to avoid a circular import
+    from comtypes.typeinfo import IRecordInfo
+
+    _SafeArrayGetRecordInfo = _oleaut32.SafeArrayGetRecordInfo
+    _SafeArrayGetRecordInfo.restype = HRESULT
+    _SafeArrayGetRecordInfo.argtypes = [POINTER(SAFEARRAY), POINTER(POINTER(IRecordInfo))]
+
+    result = POINTER(IRecordInfo)()
+    _SafeArrayGetRecordInfo(pa, byref(result))
+    return result.value
+
 SafeArrayGetElement = _oleaut32.SafeArrayGetElement
 SafeArrayGetElement.restype = HRESULT
 SafeArrayGetElement.argtypes = [POINTER(SAFEARRAY), POINTER(LONG), c_void_p]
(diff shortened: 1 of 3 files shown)


mementum:
This is for issue #63

cfarrow:
This looks good, but I haven't tested it. Do you know of any readily available typelibs that we can use to test this?

mementum:
The changes were tested against VisualChart (www.visualchart.com) which is freely available for EndOfDay data. The events take arrays of records and I decide to undertake it once I saw it was still unsupported. The source code produced by the generator is very good for the typelib.

TNick:
These changes alone are not enough to solve [the issue](https://github.com/reclosedev/pyautocad/issues/6) reported in pyautocad with `GetAttributes` where the type of oibect is `VT_DISPATCH | VT_ARRAY`.

However, taking inspiration from [here[(http://osdir.com/ml/python.comtypes.user/2008-08/msg00009.html) I un-commented lines 896 and 897 and I have access to the attributes as follows:

```python
        for att in blk.GetAttributes():
            att = comtypes.client.GetBestInterface(att)
            att_name = att.TagString
            if att_name == "ATT_1":
                att.TextString = "att_1_value"
            else:
                LOGGER.debug('Unknown attribute %s', att_name)
```

mementum:
The commit above and uncommenting VT_DISPATCH and VT_UNKNOWN do not travel togeher.

The commit above addresses a case in which even having VT_RECORD in the cvar_to_type dictionary doesn't help, because extra processing is needed to map the objects in the array to the right type, which has to be sought in the array itself.

Uncommenting those 2 lines adds 2 types to the aforementioned dictionary and according to the referenced mail exchange (from 2008) it works (like if it were a bool type for example), but "Thomas" didn't know if the references were being properly managed in those cases. Which I guess it is why the lines were left as comments.

Back to 2016 I guess if the references are really not properly managed, someone using the new enabled functionality can report a bug.

But if those 2 lines are left as comments, the functionality cannot even be used.

TNick:
Right, my bad. I can see that commenting out the changes made by this commit has no effect on the result.
As a first time user I feel shy about reporting a bug; my understanding about what the code does or should do is minimal.

junkmd:
@mementum 

This is a PR that has been posted for a long time, but no conflicts have occurred and it still appears to be useful.

However, same as @cfarrow, I am concerned about the testing method.

If you are still interested in anything related to this, please let us know what your opinions are.

PR #105, opened 2016-04-09
Stop spurious "Generating ..." output and make it part of logger.debug

mementum:


cfarrow:
Thanks. It looks like the declaration of `__verbose__` can be removed. With that change, this is good to merge.

junkmd:
@mementum 
Hi.
I have very recently become a collaborator of `comtypes`.

A long time has passed since this PR was posted.
You would have wanted to change the way to display `# Generating ...` messages from using `print` to `logger`.
They have been achieved by other PRs!

https://github.com/enthought/comtypes/blob/0f3cf2b6d309d887eed92dd2b6d4393883ccbc69/comtypes/client/_generate.py#L224-L246

https://github.com/enthought/comtypes/blob/0f3cf2b6d309d887eed92dd2b6d4393883ccbc69/comtypes/client/_generate.py#L203-L221

These changes might have been influenced by your proposal.
Thank you very much for your contributions.

I would be happy to work on #104 instead of this.

PR #116, opened 2017-03-13
Self heal

### LINKED ISSUE #114: Partial wrapper files generated in gen folder
__With Comtypes 1.1.2 and Python 2.7.10__

I have noticed that some times the generated wrappers have only the first two lines as below:

    # -*- coding: mbcs -*-
    typelib_path = u'C:\\XXXXXXXXXXXXXXXXXXXXX.tlb'

This manifests as issues such as:
`AttributeError: 'module' object has no attribute 'XXXXX'`

I had to delete the generated file and regenerate it to fix the issue whenever I see them. However I wasn't able to isolate the root cause and neither is the issue very frequent. After going through the github issues and searching over the internet, I don't see many complaining about such an issue. Only two instance that were discussed elsewhere seem relevant: 
 - https://geonet.esri.com/thread/122158
 - http://gis.stackexchange.com/questions/135741/debugging-module-object-has-no-attribute-sdeworkspacefactory

so I guess this is not specific to my setup but a generic issue.

vijairaj:
Fixes #114

### DIFF
--- a/comtypes/client/_generate.py
+++ b/comtypes/client/_generate.py
@@ -3,6 +3,7 @@ import os
 import sys
 import comtypes.client
 import comtypes.tools.codegenerator
+import imp
 
 import logging
 logger = logging.getLogger(__name__)
@@ -23,7 +24,11 @@ def _my_import(fullname):
     if comtypes.client.gen_dir \
            and comtypes.client.gen_dir not in comtypes.gen.__path__:
         comtypes.gen.__path__.append(comtypes.client.gen_dir)
-    return __import__(fullname, globals(), locals(), ['DUMMY'])
+
+    mod = imp.reload(eval(fullname)) if fullname in sys.modules \
+            else __import__(fullname, globals(), locals(), ['DUMMY'])
+    mod._comtypes_validate_file()
+    return mod
 
 def _name_module(tlib):
     # Determine the name of a typelib wrapper module.
@@ -152,8 +157,10 @@ def _CreateWrapper(tlib, pathname=None):
     # helper which creates and imports the real typelib wrapper module.
     fullname = _name_module(tlib)
     try:
-        return sys.modules[fullname]
-    except KeyError:
+        mod = sys.modules[fullname]
+        mod._comtypes_validate_file()
+        return mod
+    except (KeyError, AttributeError):
         pass
 
     modname = fullname.split(".")[-1]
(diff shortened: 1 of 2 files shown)


kviktor:
Hey, with this patch I have full files however for some reason import doesn't work on the first try only after that (it generates the source successfully into the cache dir but can't import them only after the fist run)

(I know this is almost a year old issue and maybe it is not even related to it ...)

vijairaj:
@kviktor, is the issue consistent? Can you please add more details describing the failure seen during the first time import.

junkmd:
#300 prevents partial files from being generated.

Processing order has changed.

previously: open file -> write header -> parse type library -> generate code -> write body and footer -> close file
currently: parse type library -> generate code including header and footer -> open file -> write code -> close file

The interval between opening a file, writing code, and closing it is much shorter, so "an interruption in between" is much rarer.

vijairaj:
@junkmd -- While the changed order definitely makes the issue very rare but the self heal mechanism can act as 2nd level of protection against the small odds or when the file is corrupted later.

junkmd:
@vijairaj
Certainly, importing from an empty or partial `.py` file can still occur in this situation, so I agree with the inclusion of `self heal` logic.
However, a long time has passed since this PR was first posted, the syntax has become Python 3 compliant, and the design of the `codegenerator` has changed.
If we were to reintroduce it now, what would be the best way to implement it?

junkmd:
@vijairaj 

While I am waiting for a review of the PR I have currently submitted, I thought about what would happen if I applied your proposed logic to the current `master` branch.

This code worked at runtime, but there are still problems.
https://github.com/enthought/comtypes/compare/944e279..2669789

- The code calling `_comtypes_validate_file` in the module you are importing is written in a strange position in the module.
image
image

- To solve the import failing problem in my env(is related to mentioned by @kviktor(https://github.com/enthought/comtypes/pull/116#issuecomment-355617346)?), instead of calling `_comtypes_validate_file` in `_my_import`, calling `_comtypes_validate_file` within trying to create or import the wrapper module.

- `_comtypes_validate_file` definition is not `PEP8`-compliant.

- Tests have not yet been created. I haven't figured out how to prepare the "partial module file" and how to see repaired it.

What is your opinion?

junkmd:
@vijairaj 
@kviktor 

As mentioned in #114, good suggestion about the ability to restore partial modules.

However, it is difficult to implement and also difficult to validate with `unittest`.
As for testing, if only Python3 is supported, there is `unittest.mock`, so it could be done by patching `sys.modules` and import-related processes.

And the current code and module generation process design has changed significantly since this PR was submitted, causing conflicts with the proposed code.

Due to no recent activity, I close this.
If you wish to add this feature, please rebase, resolve conflicts and re-open or submit a new PR.

PR #117, opened 2017-03-13
Embed timestamp of tlb used for cache gen.

### LINKED ISSUE #115: Support cache regeneration when switching b/w typelibs older than the comtypes cache
_(This is rather an odd requirement so I don't mind if this is not supported.)_

I have to work with comm dlls which don't use the version numbers at all, To be more specific, the version numbers are _always_ hardcoded as 1.0 and newer version's of the dlls (for reasons not under my control) don't update their version numbers. So, If I switch from a recent version of the dll/tlb to an older version, the comtypes cache is not regenerated. So I suggest to compare the timestamp of the current tlb vs the timestamp of tlb used for code generation while determining if the cache has to be re-generated instead of the current practice of comparing the time stamps of the current tlb vs the cache.

vijairaj:
Use the embedded timestamp for cache invalidation.
Fixes #115

### DIFF
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -23,21 +23,19 @@ from ctypes import *
 from _ctypes import COMError
 from comtypes import patcher
 
-def _check_version(actual):
+def _check_version(actual, tlib_cached_mtime=None):
     from comtypes.tools.codegenerator import version as required
     if actual != required:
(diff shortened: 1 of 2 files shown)


hansonap:
I think this is value added as this situation is unfortunately all too common. Many API dlls I have to work with are always versioned at 1.0.

PR #119, opened 2017-03-21
Use the logging module instead of print statements in _generate.py

orf:
It uses `logger` everywhere else, so unsure as to why it uses print in these places?


cfarrow:
LGTM

PR #121, opened 2017-03-21
Add initial appveyor file

orf:
Refs #120

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 image: Visual Studio 2015
-
+build: off
 
 init:
   - git config --global core.autocrlf input
@@ -22,4 +22,4 @@ environment:
      - py: Python36-x64
 
 test_script:
-   - C:\%py%\Scripts\pip.exe install .
+   - C:\%py%\python.exe setup.py install
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,6 @@
 image: Visual Studio 2015
 build: off
+max_jobs: 3
 
 init:
   - git config --global core.autocrlf input
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,6 @@
 image: Visual Studio 2017
 
+
 init:
   - git config --global core.autocrlf input
   
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,4 +22,4 @@ environment:
      - py: Python36-x64
 
 test_script:
-   - C:\%py%\python.exe setup.py install
+   - C:\%py%\Scripts\pip.exe install .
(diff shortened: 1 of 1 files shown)


orf:
Hmm, this should be starting a build now...

cfarrow:
I'm looking into it.

cfarrow:
It ran. I might have the integration wrong at the github end.

screen shot 2017-03-21 at 11 57 41 am

cfarrow:
It looks like notifications are configured in the appveyor.yaml file. https://www.appveyor.com/docs/notifications/#github-pull-request

cfarrow:
No, I'm misunderstanding something. That build is for master and notifications only need to be configured for custom templates. Still looking...

orf:
I think that is only for the comment, for example in [this appveyor.yml file](https://github.com/ember-cli/ember-cli/blob/master/appveyor.yml) there are no notifications, but merge requests still get the notification like so: https://github.com/ember-cli/ember-cli/pull/6708

cfarrow:
I've enabled the webhook. I don't see a way to trigger a branch build from the appveyor interface. Please push another commit.

orf:
That's the ticket! Awesome, I'll have a fiddle tomorrow and see if we can get at least the installation rather than the building passing.

cfarrow:
Cool. Thanks for taking the initiative on this!

orf:
No problem! It seems comtypes uses a VS 2010 build, but the minimum VS that appveyor offers is 2013. Is there any technical reason the project couldn't be updated to 2013?

cfarrow:
No part of the comtypes library actually needs VS. It's there to build a COM server that is used to test that functionality.

orf:
Hmm, my test was simply running `setup.py install` - should it be building the test COM server in that case? I've changed it to a `pip install` so lets see if that helps.

orf:
Oh, my bad, it appears that AppVeyor triggers the build automatically.

orf:
Hey @cfarrow - this MR is ready for merging if you're happy (you might want to let github squash it when you do).

I tried to get the tests to run as well but failed. I think they are broken and need a lot of work to get them running, but for now just running `setup.py install` on the various python versions is a win.

So as I see it the tests appear to run on Python 2.7 32bit only only. 2.6 runs, but fails due to a missing testcase method (https://ci.appveyor.com/project/cfarrow/comtypes/build/1.0.9/job/6uxqxeleqe7j0hgi). All 3.x cases fail with an `AttributeError` (https://ci.appveyor.com/project/cfarrow/comtypes/build/1.0.9/job/i528o3q1kwvqdt1v).

It seems the tests in general are in a poor state, with a custom test loader that seems pretty old. I'd quite like to convert them to `pytest` if you're willing?

cfarrow:
We can merge this as-is. I may record the test failures as their own issues, depending on how long they persist. I've moved the pytest discussion to #123.

cfarrow:
Thanks for setting this up!

Issue #102, opened 2016-04-06
TypeError: '_NamespacePath' object does not support indexing


orf:
Hello,
I've encountered a small issue with comtypes. If the `gen` folder is made by an administrator (or the user running comtypes doesn't have permissions) it fails with the exception in the issue title.

The error occurs on this line: https://github.com/enthought/comtypes/blob/master/comtypes/client/_code_cache.py#L110, after being called from the `_find_gen_dir` function (https://github.com/enthought/comtypes/blob/master/comtypes/client/_code_cache.py#L30).

This is on Python 3.4.3

cfarrow:
It looks like `path[0]` in line 110 is the culprit. This works for python versions before 3.3, but not with 3.4.3. It should not be difficult to patch, though I welcome you to submit a patch for the fastest turnaround time.

PR #132, opened 2017-09-26
fix issue #102

mortbauer:


mortbauer:
sorry that actually doesn't fix the problem, actually for me it works if the directory 'comtypes.gen' is existing and contains a file '__init__.py'

PR #134, opened 2017-11-16
Fixed issue #102

jornb:
This PR contains commit https://github.com/mortbauer/comtypes/commit/23469f5ed6ee54c281f8326fa321084e38b5a30f from @mortbauer to fix issue #102

### DIFF
--- a/comtypes/client/_code_cache.py
+++ b/comtypes/client/_code_cache.py
@@ -55,7 +55,7 @@ def _find_gen_dir():
             logger.info("Creating writeable comtypes cache directory: '%s'", gen_dir)
             os.makedirs(gen_dir)
         gen.__path__.append(gen_dir)
-    result = os.path.abspath(gen.__path__[-1])
+    result = os.path.abspath(list(gen.__path__)[-1])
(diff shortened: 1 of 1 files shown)


jornb:
@cfarrow @jaraco can anyone have a look at this? Seems comtypes is unusable for Python 3 at the moment.

jaraco:
Based on my experience with module `__path__` attributes, from work on Setuptools, this patch looks suitable. But thinking about the next person to look at the code, it may not be clear why these `list` calls are there, and one may be tempted to optimize them away. I think it would be worthwhile to have a test or a documented function or at least a comment explaining what's happening.

I think a test may prove difficult (given that the failure condition requires some external conditions), so I'll instead opt for the documented function...

jaraco:
But as I look into it more, I don't yet understand how it is that `gen.__path__` is getting to be a `_NamespacePath`. I would only expect that to happen if the `gen` package came from a folder with no `__init__.py`. But clearly the code tries to create that `__init__.py` unconditionally.

jaraco:
See #137 for an alternate approach.

PR #137, opened 2017-11-29
Ensure that gen.__path__ items can be indexed.

jaraco:
This PR supersedes #134, adapting the proposal there to include some documentation about the purpose of the `list`. Also, calls it exactly once and re-uses that value (isolating the change to a single function).


jaraco:
@jornb: Would you try out the changes and confirm they work and address the condition?

PR #140, opened 2018-01-07
fix "List index out of range" bug

totaam:
If we landed in this branch because there were no paths, we must add one to be able to call "gen_path[-1]"


stphnlyd:
++

cfarrow:
LGTM

PR #143, opened 2018-01-23
Fix crash or wrong path when using memory-only comtypes.gen

morganwillcock:
Following changes in #137 the path list for the gen cache is not updated before a value is read, so this either crashes trying to index an empty list or gives you an incorrect path. I think this is probably just a missed change in variable name, but it is breaking anything using a memory only cache (explicitly sets the path to an empty list) or a cache in a temp location. As an example, here is what happens with PyInstaller.

Current behaviour:
<pre>_create_comtypes_gen_package	INFO	Imported existing <module 'comtypes.gen' from 'C:\\Users\\MWILLC~1\\AppData\\Local\\Temp\\12\\_MEI143962\\comtypes\\gen\\__init__.pyc'>
_is_writeable	DEBUG	Path is unwriteable: [Errno 2] No such file or directory: 'C:\\Users\\MWILLC~1\\AppData\\Local\\Temp\\12\\_MEI143962\\comtypes\\gen\\tmppr51znv2'
_find_gen_dir	INFO	Creating writeable comtypes cache directory: 'C:\Users\MWILLC~1\AppData\Local\Temp\12\comtypes_cache\overseer-36'
_find_gen_dir	INFO	Using writeable comtypes cache directory: 'C:\Users\MWILLC~1\AppData\Local\Temp\12\_MEI143962\comtypes\gen'</pre>

... so it makes a path and then doesn't use it, resulting in:<pre>FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\MWILLC~1\\AppData\\Local\\Temp\\12\\_MEI143962\\comtypes\\gen\\_50A7E9B0_70EF_11D1_B75A_00A0C90564FE_0_1_0.py'</pre>

New behaviour:
<pre>_create_comtypes_gen_package	INFO	Imported existing <module 'comtypes.gen' from 'C:\\Users\\MWILLC~1\\AppData\\Local\\Temp\\12\\_MEI61722\\comtypes\\gen\\__init__.pyc'>
_is_writeable	DEBUG	Path is unwriteable: [Errno 2] No such file or directory: 'C:\\Users\\MWILLC~1\\AppData\\Local\\Temp\\12\\_MEI61722\\comtypes\\gen\\tmp3hty8xpl'
_find_gen_dir	INFO	Creating writeable comtypes cache directory: 'C:\Users\MWILLC~1\AppData\Local\Temp\12\comtypes_cache\overseer-36'
_find_gen_dir	INFO	Using writeable comtypes cache directory: 'C:\Users\MWILLC~1\AppData\Local\Temp\12\comtypes_cache\overseer-36'</pre>

### DIFF
--- a/comtypes/client/_code_cache.py
+++ b/comtypes/client/_code_cache.py
@@ -67,7 +67,7 @@ def _find_gen_dir():
         if not os.path.exists(gen_dir):
             logger.info("Creating writeable comtypes cache directory: '%s'", gen_dir)
             os.makedirs(gen_dir)
-        gen.__path__.append(gen_dir)
+        gen_path.append(gen_dir)
     result = os.path.abspath(gen_path[-1])
     logger.info("Using writeable comtypes cache directory: '%s'", result)
     return result


totaam:
Looks like you just created a duplicate of my pull request:
https://github.com/enthought/comtypes/pull/140/commits/e287210c04843e81e6fd187a7c29956142ed2be1

Looking at this again, maybe the first append should be kept too?

morganwillcock:
Ah right, I was initially searching for issues with path errors rather than a list error.
>Looking at this again, maybe the first append should be kept too?

Which line do you mean?

totaam:
> Which line do you mean?

I mean keeping both of these lines:
```
gen.__path__.append(gen_dir)
gen_path.append(gen_dir)
```

vasily-v-ryabov:
Merged the same in #148 and #140.

PR #146, opened 2018-03-18
fix gen_path empty when created with py2exe or pyinstaller

kongxinchi:


vasily-v-ryabov:
Already merged #148 and #140.

Issue #159, opened 2018-07-22
PumpEvents didn't work while PumpWaitingMessages was fine


be4pro:
```
from comtypes.client import CreateObject
import comtypes.gen.someCOM as AA
BB = CreateObject(AA.someLib)

class someLibEvents:
    def OnSth(self):
        print("message")

from comtypes.client import GetEvents
theEvent = someLibEvents()
theEventHandler = GetEvents(BB, theEvent)
```

`pythoncom.PumpWaitingMessages()` prints "message" as expected at some point

`PumpEvents(30)` shows as below
Traceback (most recent call last):
  File ".\run_xx.py", line 59, in <module>
    PumpEvents(30)
  File "D:\Python\site-packages\comtypes\client\_events.py", line 277, in PumpEvents
    ctypes.byref(ctypes.c_ulong()))
  File "_ctypes/callproc.c", line 918, in GetResult
OSError: [WinError -2147417835]

jcsteh:
Duplicate of #138. Fixed in #187 (merged).

PR #171, opened 2018-09-18
Correctly ignore error from CoWaitForMultipleHandles().

### LINKED ISSUE #138: WinError on PumpEvents timeout
The CoWaitForMultipleHandles call in the PumpEvents function throws a WindowsError on timeout. Currently, the except statement checks details.args[0] against RPC_S_CALLPENDING (previously defined as -2147417835, which is the winerror number). Details.args[0] contains the "errno". The number we're looking for is in details.args[3], but I think it might be better to refer to details.winerror instead. 

I think it should read: 
```
    try:
        try:
            res = ctypes.oledll.ole32.CoWaitForMultipleHandles(0,
                                                               int(timeout * 1000),
                                                               len(handles), handles,
                                                               ctypes.byref(ctypes.c_ulong()))
        except WindowsError as details:
            if details.winerror != RPC_S_CALLPENDING: # timeout expired
                raise
        else:
            raise KeyboardInterrupt
    finally:
        ctypes.windll.kernel32.CloseHandle(hevt)
        ctypes.windll.kernel32.SetConsoleCtrlHandler(HandlerRoutine, 0)
```

I'm pretty new to all this, but that is how I got it to work in my code.

chrullrich:
The RPC_S_CALL_PENDING error is in args[3] aka winerror.

Fixes #138, #159.

### DIFF
--- a/comtypes/client/_events.py
+++ b/comtypes/client/_events.py
@@ -276,7 +276,7 @@ def PumpEvents(timeout):
                                                                len(handles), handles,
                                                                ctypes.byref(ctypes.c_ulong()))
         except WindowsError, details:
-            if details.args[0] != RPC_S_CALLPENDING: # timeout expired
+            if details.winerror != RPC_S_CALLPENDING: # timeout expired
                 raise
         else:
             raise KeyboardInterrupt


jcsteh:
This same fix (from a different contributor) was merged in #187.

PR #187, opened 2019-07-01
fix invalid handling of WindowsError on python 3

### LINKED ISSUE #186: comtypes._comobject.winerror raises "TypeError: 'OSError' object is not subscriptable" on Python 3.7
## Steps to reproduce:
Try the following on a Python 2 and 3 interpretter:

```
import ctypes, comtypes
ex = ctypes.WinError(2)
comtypes._comobject.winerror(ex)
```

## Result
On python 2, this returns 2. ON Python 3, the following error is raised.

```
  File "comtypes\_comobject.py", line 51, in winerror
    code = exc[0]
TypeError: 'FileNotFoundError' object is not subscriptable
```

leonardder:
fixes comtypes._comobj.winerror by no longer subscripting the exception. This pull requests makes this function use instance checking instead, raising a TypeError when the instance is neither COMError nor WindowsError.

Fixes #186

It also fixes comtypes.client.PumpEvents raising RPC_S_CALLPENDING.

### DIFF
--- a/comtypes/client/_events.py
+++ b/comtypes/client/_events.py
@@ -275,8 +275,8 @@ def PumpEvents(timeout):
                                                                int(timeout * 1000),
                                                                len(handles), handles,
                                                                ctypes.byref(ctypes.c_ulong()))
-        except WindowsError, details:
-            if details.args[0] != RPC_S_CALLPENDING: # timeout expired
+        except WindowsError as details:
+            if details.winerror != RPC_S_CALLPENDING: # timeout expired
                 raise
         else:
             raise KeyboardInterrupt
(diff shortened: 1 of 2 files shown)


comtypes/client/_events.py
```diff
@@ -275,8 +275,8 @@ def HandlerRoutine(dwCtrlType):
                                                                int(timeout * 1000),
                                                                len(handles), handles,
                                                                ctypes.byref(ctypes.c_ulong()))
-        except WindowsError, details:
-            if details.args[0] != RPC_S_CALLPENDING: # timeout expired
+        except WindowsError as details:
```
vasily-v-ryabov: If you change comma to `as` keyword, then you need to change it everywhere to keep consistency (not sure it's time for big re-factoring). `setup.py` script uses 2to3 fixer at the moment. So keeping comma wouldn't be a problem. With this small change I'd merge this PR without risk to have concerns from Enthought's colleagues. Thanks for understanding.

leonardder:
I'm happy to revert the 'as', no problem at all.

leonardder:
@vasily-v-ryabov: Your request has been applied.

PR #189, opened 2019-07-30
Sets zip_safe flag in setup.py

kdschlosser:
without the zip_safe set to False when setuptools installs the library
either using easy_install, setup_requires or install_requires these warnings
get produced. The issue lies in the use of a CI service like Appveyor, the
warnings are output using stderr and this triggers the build to fail.

```python
zip_safe flag not set; analyzing archive contents...
comtypes.__init__: module references __file__
comtypes.client._code_cache: module references __path__
comtypes.client._generate: module references __file__
comtypes.client._generate: module references __path__
comtypes.server.register: module references __file__
comtypes.test.TestComServer: module references __file__
comtypes.test.TestDispServer: module references __file__
comtypes.test.__init__: module references __path__
comtypes.test.test_dispinterface: module references __file__
comtypes.test.test_findgendir: module references __path__
comtypes.test.test_safearray: module references __file__
comtypes.test.test_server: module references __file__
comtypes.test.test_urlhistory: module references __file__
comtypes.test.test_win32com_interop: module references __file__
comtypes.tools.codegenerator: module references __path__
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -167,6 +167,7 @@ setup_params = dict(
         "comtypes.tools",
         "comtypes.test",
     ],
+    zip_safe=False
 )
 
 if __name__ == '__main__':


kdschlosser:
as a note I would imagine that because of generated files being in the libraries path that this would also make it not zip safe.

setup.py
```diff
@@ -167,6 +167,7 @@ def run(self):
         "comtypes.tools",
         "comtypes.test",
     ],
+    zip_safe=False
```
nanonyme, line 170: This is a bug if this doesn't work. According to its own documentation comtypes is zip_safe and generates modules in memory in that case. If this project drops that feature, then it needs to be properly removed from codebase and documentation

nanonyme (changes requested):

setup.py
kdschlosser: I believe that comtypes always generates files into a folder. I have never seen it generate the files into memory. I have a project that I work on where the application gets compiled using py2exe. The libraries the application uses get zipped up into a single zip file. I have to explicitly set comtypes.gen.\_\_path\_\_ to a folder that I have made on the users system for storage of these generated files. Otherwise comtypes will not run. I do not remember the exact error that happens I would have to change the application and compile it to get the error it produces.

I also do not recall seeing any Voodoo magic code for creating modules that get stored in the generation folder. I could have missed it tho.

kdschlosser: So now that I am thinking about it, generated files should not be stored in the libraries path any how. They should be stored in the users application data or in the program data. This would be %appdata% (c:\users\{user name}\appdata\roaming) or %programdata% (c:\programdata). The generated modules do not need to be located in the comtypes.gen folder. the path to where the generated files actually exist needs to be set in comtypes.gen.\_\_path\_\_. This allows comtypes to function properly.

There can also potentially be another issue with writing the files to a folder in the library path.
The users have the ability to install Python into the program files directory. The thing is that with Windows 10 there can be permissions issues with an application creating files in the program files directory or any of it's sub folders.  So writing the generated files to either program data or app data is going to kill 2 birds with a single stone. It will solve the zip_safe issue and it will also solve a potential permissions problem on Windows 10.

nanonyme: From https://pythonhosted.org/comtypes/
```


    This variable determines the directory where the typelib wrappers are written to. If it is None, modules are only generated in memory.

    comtypes.client.gen_dir is calculated when the comtypes.client module is first imported. It is set to the directory of the comtypes.gen package when this is a valid file system path; otherwise it is set to None.

    In a script frozen with py2exe the directory of comtypes.gen is somewhere in a zip-archive, gen_dir is None, and even if tyelib wrappers are created at runtime no attempt is made to write them to the file system. Instead, the modules are generated only in memory.

    comtypes.client.gen_dir can also be set to None to prevent writing typelib wrappers to the file system. The downside is that for large type libraries the code generation can take some time.
```

nanonyme: I have actually tested it that comtypes.client.gen_dir is set to None as the first thing, then modules in reality were generated into memory. The documentation talks *exactly* of py2exe. If this functionality doesn't work, then that's a bug.

nanonyme: The exact logic is here https://github.com/enthought/comtypes/blob/master/comtypes/client/_code_cache.py#L54-L61

IMHO this is overly complex. It should probably just always fallback to generating into memory with py2exe.

nanonyme: I'm not overly sure the implementation is anymore according to documentation at this point. https://github.com/enthought/comtypes/blob/master/comtypes/client/_code_cache.py#L53-L64

Documentation claims in those last two cases return value is None. Seems at some point this stopped happening.

nanonyme: Let's continue the discussion in the issue I created. comtypes already works similar to how you suggested (against its documentation) that it uses user appdata directory in non-py2exe case and tempfile.gettempdir() in py2exe case.

PR #211, opened 2020-06-16
Remove Windows CE support

forderud:
I don't think the Windows CE kernel is supported by Microsoft any more. The latest release was 8.0 in 2013 according to Wikipedia (https://en.wikipedia.org/wiki/Windows_Embedded_Compact).

Python 3.6 and newer also no longer support Windows CE (https://docs.python.org/3.6/library/os.html#os.name).

I recommend to "Hide whitespace changes" when reviewing this PR.
image

### DIFF
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -148,10 +148,7 @@ def CoInitialize():
 
 def CoInitializeEx(flags=None):
     if flags is None:
-        if os.name == "ce":
-            flags = getattr(sys, "coinit_flags", COINIT_MULTITHREADED)
-        else:
-            flags = getattr(sys, "coinit_flags", COINIT_APARTMENTTHREADED)
+        flags = getattr(sys, "coinit_flags", COINIT_APARTMENTTHREADED)
     logger.debug("CoInitializeEx(None, %s)", flags)
     _ole32.CoInitializeEx(None, flags)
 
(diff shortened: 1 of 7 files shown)


vasily-v-ryabov:
Not sure this removal makes much sense. @cfarrow @jaraco are you aware about existing WinCE users?

jaraco:
I'm not aware of any WinCE users, but that's not saying much. I agree it would be worthwhile to remove crufty code or at the very least abstract it away so it's not interleaved through the code.

> Python 3.6 and newer also no longer support Windows CE (https://docs.python.org/3.6/library/os.html#os.name).

Is there a better reference describing the dropped support for Windows CE? Absence of a registered `os.name` seems inadequate.

I'm inclined to agree that it's unlikely this code is being used (or that CE users couldn't readily pin to older releases). Unless there's more information about active CE users, I'd recommend removing the functionality (with a backward-incompatible indication), but if there's a meaningful uproar as a result, consider reverting the change and restoring support.

forderud:
Quote from https://bugs.python.org/issue27355 :
> Windows CE support never survived the transition to Python 3.  It has clearly never even been compiled since Python 3.0 shipped.

cfarrow:
I'm on the same page as @jaraco.

The risk of removing it is very low. The worst-case scenario is that a small number of users are inconvenienced and we learn who is using WinCE. The best case is that we go back to not worrying about who is using WinCE. :)

wkschwartz:
README still says CE is supported (I don't use CE, just happened to notice it)

PR #172, opened 2018-10-23
Fix importing newly-created modules on Python 3

fxthomas:
Python 3's import mechanism caches the contents of a directory
containing modules (such as comtypes.gen) and relies on the "modified
time" (mtime) attribute to determine if it needs to reload that
directory.

In Windows, the mtime is sometimes not updated very often when creating
modules in quick succession, which means that the mtime stays identical,
which then appears in comtype's case as random module import failures
when using a new COM object for the first time.

See https://stackoverflow.com/questions/52933869/why-does-importing-fails-after-creating-a-module-under-python-3-on-windows for a more detailed explanation.

(Fixes #161)

PR #214, opened 2020-07-02
BUILD: End `bdist_wininst` support in `setup.py`

### LINKED ISSUE #191: `pip install comtypes` fails
On Windows 10, x86-64 bit, Python 3.7.4. I think this is actually the bug referenced here: https://github.com/pywinauto/pywinauto/issues/800

Output I'm seeing:

```
    copying build\scripts-3.7\clear_comtypes_cache.py -> C:\Users\WolfgangRichter\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.7_qbz5n2kfra8p0\LocalCache\local-packages\Python37\Scripts
    error: error in setup script: command 'bdist_wininst' has no such option 'install_script'
    ----------------------------------------
ERROR: Command errored out with exit status 1: 'C:\Users\WolfgangRichter\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.7_qbz5n2kfra8p0\python.exe' -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\WOLFGA~1\\AppData\\Local\\Temp\\pip-install-ngcm1q34\\comtypes\\setup.py'"'"'; __file__='"'"'C:\\Users\\WOLFGA~1\\AppData\\Local\\Temp\\pip-install-ngcm1q34\\comtypes\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record 'C:\Users\WOLFGA~1\AppData\Local\Temp\pip-record-g4ot3blk\install-record.txt' --single-version-externally-managed --compile --user --prefix= Check the logs for full command output.
```

### LINKED ISSUE #196: Problem installing comtypes on Windows 10 Pro when running pip 19.3.1
I'm getting the following error:

     error: error in setup script: command 'bdist_wininst' has no such option 'install_script'

when running

   pip install comtypes

using pip 19.3.1 on windows 10 Pro 10.0.19536 N/A Build 19536

### LINKED ISSUE #199: error in setup script: command 'bdist_wininst' has no such option 'install_script'
I have this error when trying pip install com types 
system windows 10 64
python :  3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 01:54:44)

Error;
 error: error in setup script: command 'bdist_wininst' has no such option 'install_script'

wkschwartz:
Fixes #191
Fixes #196 
Fixes #199 
Ref indygreg/PyOxidizer#202

I'll start with the post hoc justification for this change: `bdist_wininst` has been deprecated starting in Python 3.8. See
- notice:      https://docs.python.org/3/whatsnew/3.8.html#deprecated
- discussion:  https://discuss.python.org/t/deprecate-bdist-wininst/1929
- bug tracker: https://bugs.python.org/issue37481

The real reason for this change is that I can't seem to get around #199 when building a PyOxidizer package using PyOxidizer's `dist.pip_install` Skylark function. Upgrading Wheel to the latest version solves #199 for me in a regular virtual environment, but requiring the most recent Wheel version in my [`pyproject.toml`](https://www.python.org/dev/peps/pep-0518/)'s `build-system.requires` list doesn't solve #199. (Verbose output from Pip confirms the latest Wheel version is indeed installed.)

So removing `bdist_wininst` support solves my very niche problem, but because `bdist_wininst` is so rarely used and is deprecated anyway, I figure there's little harm to anyone else in removing it from comtypes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,8 +111,7 @@ class post_install(install):
 
     def run(self):
         install.run(self)
-        # Custom script we run at the end of installing - this is the same script
-        # run by bdist_wininst
+        # Custom script we run at the end of installing
         if not self.dry_run and not self.root:
             filename = os.path.join(self.install_scripts, "clear_comtypes_cache.py")
             if not os.path.isfile(filename):
@@ -124,8 +123,6 @@ class post_install(install):
             except subprocess.CalledProcessError:
                 print("Failed to run post install script!")
 
-
-options={"bdist_wininst": {"install_script": "clear_comtypes_cache.py"}}
 
 setup_params = dict(
     name="comtypes",
@@ -151,7 +148,6 @@ setup_params = dict(
     classifiers=classifiers,
 
     scripts=["clear_comtypes_cache.py"],
-    options=options,
 
     cmdclass={
         'test': test,


vasily-v-ryabov:
Hmm... Clearing comtypes cache is important "post-install" step. Can you think about getting it back another way?

wkschwartz:
Looks like `clear_comtypes_cache.py` already/still runs during installation:

https://github.com/enthought/comtypes/blob/47517d697aaa0b4f6abfd25a081e6e5c764b1d17/setup.py#L112-L124

I believe that the code I removed only affects when someone runs `python setup.py bdist_wininst`, not any other of `setup.py`'s commands.

vasily-v-ryabov:
OK, that makes sense. Thanks for the diving into the problem!

wkschwartz:
I need this PR, but since the next release won't come until [early 2021](https://github.com/enthought/comtypes/pull/220#issuecomment-732770622) and I'm using [Pip's hash-checking mode](https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode) (which doesn't support installing from Git), I need an archive somewhere accessible. I'm putting it here since its main value is that it contains the patch in this PR. This way if anyone else needs it, they can find it. It's 8d67890 so it also contains #215, #172, and #220.
- [comtypes-1.1.7-19-g8d67890.tar.gz](https://github.com/enthought/comtypes/files/5599400/comtypes-1.1.7-19-g8d67890.tar.gz) `--hash=sha256:acfe45908f61a20760f391fa63dbfc218ef565d0540cd11b5eb617830da43bf2`

For the avoidance of doubt: this is **not an official release**.

PR #215, opened 2020-07-23
Allow customization of shutdown behavior

lassoan:
Instead of deleting comtypes.shutdown it is made private (by renaming to comtypes._shutdown).
This allows applications to unregister shutdown from atexit and call it at the appropriate time instead.

It fixes crash during exit in applications that embed Python (and thus may require a custom shutdown sequence).


lassoan:
Thank you!

PR #223, opened 2021-01-03
Use absolute path to DLL before getting its timestamp (fix #222)

### LINKED ISSUE #222: FileNotFoundError for GetModule with relative path
Hi,

The following snippet:
```
from comtypes.client import GetModule
GetModule('UIAutomationCore.dll')
```
Results in an error with the message `FileNotFoundError: [WinError 2] The system cannot find the file specified: 'UIAutomationCore.dll'` when run using comtypes `1.1.8`, however it works correctly when downgrading back to `1.1.7`.

The traceback points to this line: https://github.com/enthought/comtypes/blob/e5eeb5eac554531943f17f62815910fc1fd16017/comtypes/tools/codegenerator.py#L261

It seems like it's missing a check for the case where the filename is `None`, like is done earlier in the same function.

EDIT: It actually might not be `None`, but a file somewhere not in the current directory, as indicated by the documentation in GetModule()

vasily-v-ryabov:
PR #117 broken loading COM DLLs from relative path. This fix is tested with `UIAutomationCore.dll` which is located in the system path.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,11 @@
 Comtypes CHANGELOG
 ==================
 
+Release 1.1.9
+-------------
+ * Fix loading module from relative path.
+ * Use comtypes release version in code generator version check.
+
 Release 1.1.8
 -------------
  * Fix No module named 'comtypes.gen.stdole'. Thanks @fxthomas.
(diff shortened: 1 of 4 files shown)


nanonyme:
While this looks fine to me, could you briefly explain what that removal of WinCE classifier was about?

vasily-v-ryabov:
Hi @nanonyme it's a trailing change after PR #211.

PR #226, opened 2021-04-23
Add quirk for Windows 7 peculiarities

nanonyme:
Possibly due to missing KB982110 QueryPathOfRegTypeLib
sometimes bugs out on Windows 7 and does not give full
path. It appears giving out mtime of 0 is less harmful
here than breaking completely

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -4,6 +4,7 @@ import os
 import cStringIO
 import keyword
 import ctypes
+import errno
 
 from comtypes.tools import typedesc
 import comtypes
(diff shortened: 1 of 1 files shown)


nanonyme:
Ping @vasily-v-ryabov this is quite hacky but it seemed to result in codegen not breaking in a heap on Windows 7. To reproduce the problem (or validate the fix) gen directory must be deleted completely.

nanonyme:
Should be Python 2.7-compliant.

vasily-v-ryabov:
I have another idea: do `GetModuleFileName` for `HMODULE` handle of the library. Maybe additional call to `LoadLibrary` is needed.

nanonyme:
If you can provide patch of alternative, I can test. The issue reproduces 100% of time on a Windows 7 image I have currently.

vasily-v-ryabov:
Please try #227. It's kind of hybrid patch with #226 and my idea. It shouldn't re-generate module cache second time. If it re-generates it for loadable DLL type lib, it means the problem still exists. If it is not loadable DLL, it's OK to re-generate the code every time.

PR #227, opened 2021-04-24
Workaround for Windows 7 bug in QueryPathOfRegTypeLib

vasily-v-ryabov:
@nanonyme please test this fix on your Windows 7, especially on both 32-bit and 64-bit Python. @eltimen confirmed it works on his Windows 7.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,10 @@
 Comtypes CHANGELOG
 ==================
 
+Release 1.1.10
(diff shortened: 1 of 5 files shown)


nanonyme:
I can no longer reproduce the error with these changes on Windows 7 32 or Windows 7 64.

vasily-v-ryabov:
Released on PyPI as 1.1.10.

PR #234, opened 2021-07-08
Fix 64bit inproc server

### LINKED ISSUE #150: Python-implemented COM servers broken with 64bit Python3
Support for implementing COM servers in Python appears to be at least partly broken when running with 64bit Python 3.6. Minimal reproducer: https://github.com/forderud/public/tree/master/comtypes-64bit .

In this reproducer, `comtypes.client.CreateObject(...)`  fails when running with 64bit Python3.6. The problem is _not_ observed when running with either 32bit Python 2.7, 64bit Python 2.7 or 32bit Python 3.6.

The problem appears to be caused by passing of a memory address as `int`. This fails with a `int too long to convert` OverflowError when calling `CopyComPointer` in 64bit.
image

I suspect that the 2nd argument to `CopyComPointer` need to be changed from `int` to a more suitable type.

klusidy:
Passing memory address as int can fail with "int too long to convert" in 64bit python.
Casting it to c_void_p fixed the problem for me. Not sure whether this is the correct place for the conversion though.
Hopefully fixes #150, see the discussion for more details

### DIFF
--- a/comtypes/server/inprocserver.py
+++ b/comtypes/server/inprocserver.py
@@ -119,7 +119,7 @@ def DllGetClassObject(rclsid, riid, ppv):
         if not cls:
             return CLASS_E_CLASSNOTAVAILABLE
 
-        result = ClassFactory(cls).IUnknown_QueryInterface(None, ctypes.pointer(iid), ppv)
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov (approved):
Thanks for the fix! It looks good to me. I just want to test it on my side before merging.

Issue #205, opened 2020-05-15
Recent numpy does not have ctypeslib._typecodes


jove1:
Hello,

I've had few problems with comtypes recently:

at first, recent [numpy](https://github.com/numpy/numpy/commit/47403aba1977882b5145bdaa6b85fbf418dc2fd3) does no longer have ctypeslib._typecodes thus safearray_as_ndarray was no longer working, moreover the true exception got masked as AttributeError somewhere in the case normalisation thus it was difficult to find, but fixed easily by monkey patching:
```
import numpy.ctypeslib
numpy.ctypeslib._typecodes = numpy.ctypeslib._scalar_type_map # changed in numpy 47403aba
```
Unfortunately, it is beyond my capability to provide proper patch for this. At least I hope that this
information will be useful.

Regards

Jozef Vesely

PS: I believe #201 has the same problem
PS2: already reported as #178

fockez:
Thank you for the information. I added the line into npsupport.py for the test and it helps.

mobiusklein:
#220 should fix this

vasily-v-ryabov:
1.1.13 with the fixes is out.

PR #235, opened 2021-08-06
fix should address issue #205

mworion:
There is a mix between ctypes and dtypes as dtypes are in values() and ctypes are in keys()


vasily-v-ryabov:
Merged in #308.

PR #239, opened 2021-09-01
fix: make numpy typecodes consistent across numpy versions

### LINKED ISSUE #238: Typo in safearray_as_ndarray conversion to numpy for _get_elements_raw()?
In the below line which is part of unpacking a safearray into (perhaps) an ndarray, the mapping of ctypes types to numpy dtypes is checked to see if the type of the safearray is supported by numpy, so that a direct conversion of the data is possible. However, it checks the values() which as far as I can tell are the numpy dtypes, surely it should be checking the keys() instead?

https://github.com/enthought/comtypes/blob/8f3abc93c213fccdf9cae54aa88bfc1dfc17b8e9/comtypes/safearray.py#L311

bennyrowland:
In newer versions of numpy the ctypeslib._typecodes variable is no longer
present, so comtypes populates it itself if it is not found. Previously the
created dictionary was a mapping of ctypes classes to numpy dtypes, whereas
the previous _typecodes dictionary mapped numpy dtype string representations
to ctypes classes. This fix rearranges the mapping in the dictionary to be
consistent with the previous format.

Closes #238

### DIFF
--- a/comtypes/npsupport.py
+++ b/comtypes/npsupport.py
@@ -86,15 +86,15 @@ def _check_ctypeslib_typecodes():
     except ImportError:
         from numpy.ctypeslib import as_ctypes_type
 
-        ctypes_to_dtypes = {}
+        dtypes_to_ctypes = {}
 
         for tp in set(np.sctypeDict.values()):
             try:
                 ctype_for = as_ctypes_type(tp)
-                ctypes_to_dtypes[ctype_for] = tp
+                dtypes_to_ctypes[np.dtype(tp).str] = ctype_for
             except NotImplementedError:
                 continue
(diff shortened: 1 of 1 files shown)


cfarrow (commented):
This looks good to me. 

We probably shouldn't be modifying the state of `ctypeslib` on line 97, but that can wait for a future PR.

@vasily-v-ryabov Would you like to take a look before this gets merged? I am unable to test it.

vasily-v-ryabov:
I'm also unable to test it. I have no such test cases that are fast to run. I need time to reproduce it and confirm it's working. If someone can confirm it works on both versions of numpy, I can rely on it and merge.

PR #241, opened 2021-11-02
Fix faulty logic - timestamping - another attempt!

fusentasticus:
Hi, may I kindly ask you for priority attention to the problems below? They have bugged the client side of comtypes for some time now. So I have to the best of my ability tried to analyze this in-depth. If I have a solution, then you'd be able to verify that it works without too much trouble I'd think.

My reading of the existing code is as follows. 

In GetModule, the pathname calculated https://github.com/enthought/comtypes/blob/8f3abc93c213fccdf9cae54aa88bfc1dfc17b8e9/comtypes/client/_generate.py#L70-L90
is absolute if the pathname can be resolved into an existing typelib on disk according to the directory of the calling module (when that directory exists). Otherwise, the pathname is relative, but may still be good if

 - it corresponds to a typelib in a file relative to cwd or 
 - it can be found with the default Windows typelib search path. 
 
This is confusing already because the two latter conditions are not separated; in both cases, a relative path is carried forward. 

But in all subcases of the outer `if` -- the initial tlib argument is a string -- the call:

https://github.com/enthought/comtypes/blob/8f3abc93c213fccdf9cae54aa88bfc1dfc17b8e9/comtypes/client/_generate.py#L90

is executed (with possible Windows exception).  [The "Ex" suffix here denotes that tlib is not being registered as a side-effect (as I understand it).]
 
In my proposed code, this distinction between the two cases above is made clearer with the flag `path_exists`; and most of this logic is moved to its own function  `_resolve_filename` for clarity.  Also, in the case of a path that corresponds to a file relative to cwd, the pathname is absolutized when carried forward. That ensures that a relative pathname is one that succeeds only by lookup in the default search path and for which no location was returned oracularly, that is, by `comtypes.tools.tlbparser.get_tlib_filename` . 

FOR THAT CASE ONLY IS THE TIMESTAMP NOT CHECKED (in the new code)! That is at least the idea.

To see this, we follow pathname. It is passed through _CreateWrapper https://github.com/enthought/comtypes/blob/8f3abc93c213fccdf9cae54aa88bfc1dfc17b8e9/comtypes/client/_generate.py#L159 to generate_module https://github.com/enthought/comtypes/blob/8f3abc93c213fccdf9cae54aa88bfc1dfc17b8e9/comtypes/tools/tlbparser.py#L736 to generate_code https://github.com/enthought/comtypes/blob/8f3abc93c213fccdf9cae54aa88bfc1dfc17b8e9/comtypes/tools/codegenerator.py#L215 where the current logic is https://github.com/enthought/comtypes/blob/8f3abc93c213fccdf9cae54aa88bfc1dfc17b8e9/comtypes/tools/codegenerator.py#L264-L274 

Thus,  the generated code will compare the future timestamp of pathname, now called filename, to the current timestamp, acquired by an os.stat on the filename. 

This code is problematic or wrong for several reasons I think:
- trivially, when tlib_mtime is None, the %-formatting in L274 fails, because None is not an "f", a float in %-formatting -- not good!
- the LoadTypeLib should be LoadTypeLibEx -- as it was in GetModule where the call was issued as LoadTypeLibEx (so as not to register) [I think]
- there is really no need to issue the call again I think
- if the typelib is not registered, then the subsequent call to get_tlib_filename will not return
  the filepath - that's another bug in my current opinion -- it means that timestamping will *not* be applied in such cases

My proposed patch should address all of these issues.

-----

So is my proposal correct? I don't know. The history here is a bad omen -- I think this PR may be the  fourth or fifth attempt at correcting the underlying unsoundness of the version control of the generated py file from the typelib file.

Also, I have not tested this except for playing around with test cases that look like this: (where I'm trying to get the tybelib of pyia2 to be wrapped)

 py -3.8-32 -c "from comtypes.client import GetModule; import logging; logging.getLogger().setLevel(logging.DEBUG); import comtypes.client._generate; comtypes.client._generate.logger.addHandler(logging.StreamHandler()); import comtypes.tools.codegenerator; comtypes.tools.codegenerator.logger.addHandler(logging.StreamHandler()); GetModule(r'C:\Users\fusen\AppData\Local\Programs\Python\Python38-32\Lib\site-packages\pyia2\ia2.tlb')"

### DIFF
--- a/comtypes/client/_generate.py
+++ b/comtypes/client/_generate.py
@@ -1,8 +1,10 @@
 import types
 import os
 import sys
+import comtypes
 import comtypes.client
 import comtypes.tools.codegenerator
+import comtypes.tools.tlbparser
 import importlib
 
 import logging
@@ -27,6 +29,36 @@ def _name_module(tlib):
                libattr.wMajorVerNum,
                libattr.wMinorVerNum)
     return "comtypes.gen." + modname
+
+def _resolve_filename(tlib_string, dirpath):
+    """Tries to make sense of a type library specified as a string.
+    
+    Args:
+        tlib_string: type library designator
+        dirpath: a directory to relativize the location
+
+    Returns:
+
+      (abspath, True) or (relpath, False), 
+    
+    where relpath is an unresolved path."""
+    assert isinstance(tlib_string, basestring)
+    # pathname of type library
+    if os.path.isabs(tlib_string):
+        # a specific location
+        return tlib_string, True
+    elif dirpath:
+        abspath = os.path.normpath(os.path.join(dirpath, tlib_string))
+        if os.path.exists(abspath):
+            return abspath, True
+    # try with respect to cwd (if _getmodule executed from command line)
+    abspath = os.path.abspath(tlib_string)
+    if os.path.exists(abspath):
+        return abspath, True
+    # Otherwise it may still be that the file is on Windows search
+    # path for typelibs, and we leave the pathname alone.
+    return tlib_string, False
+
 
 def GetModule(tlib):
     """Create a module wrapping a COM typelibrary on demand.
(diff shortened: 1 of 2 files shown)


wroy-ca:
Thank you. This proposal fixed the underlying issue I had with comtypes.
```
pip install https://github.com/fusentasticus/comtypes/archive/refs/heads/master.zip
```

vasily-v-ryabov:
Hi @fusentasticus I've rebased your fix in PR #259.
@fusentasticus @wroyca could you please test this new PR to make sure rebase was correct?

vasily-v-ryabov:
Merged in PR #259. Closing this one.

PR #252, opened 2022-01-17
Python 3 build & install support

cmin764:
Installing this **comtypes-fork** package will replace the already existing **comtypes** Py2 one with its own Py3 flavour. (doesn't matter if **comtypes** isn't installed yet)

Drawback: each time we want to upgrade to a newer version of **comtypes**, we need to maintain in-sync this fork as well and publish the matching newer version.

Wheel to test with attached below. (should fix your broken env after install -- to be added in the next version of **rpaframework**)
[comtypes_fork-1.1.10-py3-none-any.whl.zip](https://github.com/enthought/comtypes/files/7882921/comtypes_fork-1.1.10-py3-none-any.whl.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.eggs/
+build/
+*.egg-info/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .eggs/
-build/
 *.egg-info/
+build/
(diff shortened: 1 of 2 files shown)


cmin764:
Wrongly opened PR. (will open another soon)

Issue #180, opened 2019-03-27
Syntax error when trying to import comtypes on Python 3.6.7


yurgenzzz:
Sorry, I'm a totally new in Python, Git and such, so maybe that problem was solved before, and I just missed it... I need to install comtypes locally with pip. The PC is offline, Win 7 x86.
I downloaded comtypes in zip file and used `pip install comtypes-1.1.7.zip` to install it. Successfully installed, but when I try to import the module, it shows error message:
>>> import comtypes
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python367\lib\comtypes\__init__.py", line 381
except COMError, err:
               ^
SyntaxError: invalid syntax

diogofriggo:
I'm experiencing the same error on Python 3.7.10 and comtypes 1.1.10

drijonha:
I'm experiencing the same error on Python 3.9.0 and comtypes 1.1.10

Vivojay:
I'm experiencing the same error on Python 3.9.9 and comtypes 1.1.10

filipporomani:
Same for me

lededev:
`except COMError, err:`
change to
`except (COMError, err) as err:`

Vivojay:
So do we have to fork it and then download that forked git repo instead of comtypes?

lededev:
or you can use Python 3.9.

Vivojay:
I am using python 3.9.9 ...

lededev:
Anaconda's python 3.9.7 work well with comtypes.

DH-MP:
On python 3.7, workaround for me is downgrading to wheel==0.36.2 and setuptools==57.0.0
pip install setuptools==57.0.0
pip install wheel==0.36.2
then pip uninstall comtypes 
and pip install --no-cache-dir comtypes  (make sure it's not using cache)


If you are doing it in a requirement file do the setuptools and wheel seperately (another requirement file) before installing the main requirement file.
pip install --no-cache-dir -r download_setup.txt
pip install --no-cache-dir -r your_requirement.txt

PhilippSelenium:
Come on, two years and still not fix ...

Vivojay:
I was wondering the same thing...
I would really wish they made this fix...

filipporomani:
[...]

PR #253, opened 2022-01-18
Python 3 changes

kdschlosser:
since 2.x is no longer being supported build_py_2to3 in distutils has been dropped and that is the reason why on python 3.9 and above installations there is no 2 to 3 conversion done. 

This is what the 2 to 3 conversion outputs when installing on python 3.8.

### DIFF
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -372,7 +372,7 @@ class _cominterface_meta(type):
 
                     try:
                         result = self.Item(*args)
-                    except COMError, err:
+                    except COMError as err:
                         (hresult, text, details) = err.args
                         if hresult == -2147352565:  # DISP_E_BADINDEX
                             raise IndexError("invalid index")
@@ -391,7 +391,7 @@ class _cominterface_meta(type):
                     "Attempt 'self.Item[index] = value'"
                     try:
                         self.Item[index] = value
-                    except COMError, err:
+                    except COMError as err:
                         (hresult, text, details) = err.args
                         if hresult == -2147352565:  # DISP_E_BADINDEX
                             raise IndexError("invalid index")
@@ -488,7 +488,7 @@ class _cominterface_meta(type):
                 if is_prop:
                     self.__map_case__[name[5:].lower()] = name[5:]
 
-        for (name, nargs), methods in properties.items():
+        for (name, nargs), methods in list(properties.items()):
             # methods contains [propget or None, propput or None, propputref or None]
             if methods[1] is not None and methods[2] is not None:
                 # both propput and propputref.
@@ -574,7 +574,7 @@ class _cominterface_meta(type):
             for itf in self.mro()[1:-1]:
                 result += len(itf.__dict__["_methods_"])
             return result
-        except KeyError, err:
+        except KeyError as err:
             (name,) = err.args
             if name == "_methods_":
                 raise TypeError("baseinterface '%s' has no _methods_" % itf.__name__)
@@ -655,7 +655,7 @@ class _cominterface_meta(type):
                 return rescode
 
             rescode = list(rescode)
-            for outnum, o in outargs.items():
+            for outnum, o in list(outargs.items()):
                 rescode[outnum] = o.__ctypes_from_outparam__()
             return rescode
         return call_with_inout
@@ -779,7 +779,7 @@ class _cominterface_meta(type):
                     self.__map_case__[name[5:].lower()] = name[5:]
 
         # create public properties / attribute accessors
-        for (name, doc, nargs), methods in properties.items():
+        for (name, doc, nargs), methods in list(properties.items()):
             # methods contains [propget or None, propput or None, propputref or None]
             if methods[1] is not None and methods[2] is not None:
                 # both propput and propputref.
@@ -895,9 +895,8 @@ class _compointer_meta(type(c_void_p), _cominterface_meta):
     "metaclass for COM interface pointer classes"
     # no functionality, but needed to avoid a metaclass conflict
 
-class _compointer_base(c_void_p):
+class _compointer_base(c_void_p, metaclass=_compointer_meta):
     "base class for COM interface pointer classes"
-    __metaclass__ = _compointer_meta
     def __del__(self, _debug=logger.debug):
         "Release the COM refcount we own."
         if self:
@@ -1017,7 +1016,7 @@ class BSTR(_SimpleCData):
 ################################################################
 # IDL stuff
 
-class helpstring(unicode):
+class helpstring(str):
     "Specifies the helpstring for a COM method or property."
 
 class defaultvalue(object):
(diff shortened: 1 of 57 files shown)


cmin764:
Looks good, thank you! I'm curious, what's the Python 3.x version range where this code is still compatible with? Like does it support and behave the same with Python 3.7.5 interpreters?

kdschlosser:
it should work with 3.4 to current.

PhilippSelenium:
https://github.com/enthought/comtypes/issues/180

comtypes/__init__.py
```diff
@@ -488,7 +488,7 @@ def _make_dispmethods(self, methods):
                 if is_prop:
                     self.__map_case__[name[5:].lower()] = name[5:]
 
-        for (name, nargs), methods in properties.items():
+        for (name, nargs), methods in list(properties.items()):
```
dosas, line 491: If I am not mistaken the list conversion can be omitted in an iterating context. AFAIK it is even faster that way.

Thanks for porting this to python3.

kdschlosser: Well, I didn't really do a proper port.. I just did a PR of what distutils output for python 3 code. Easiest and quickest way to get it back up and running. I figured changes like the one above could be handled after this PR got merged, if it even gets merged.

The distutils 2 to 3 conversion does "literal" conversions to keep things as they would be in python 2.  `dict.items()` in python 2 returned a list ad not a generator so a proper conversion would be to wrap the call in `list()`. There are going to be changes that can be made in order to make the code more efficient and also to remove blocks of code that no longer need to be there because they are there only for python 2. 

For the sake of just getting comtypes working again I think it best to leave this PR as is and make changes afterwards and have a PR for each change. I suggest this because we know that the python 2 to 3 conversion does work without breaking anything and it's best no to introduce any changes that might break something. 

While I do know that the code change you have brought up wouldn't break anything it feel it is best to wait to make those changes.

Any thought on agreement or disagreement are  helpful  and wanted.  So anyone that want to add to this conversation please chime in. 

While I am not new  to using comtypes I have not spent the time to really go over the library to understand the mechanics of it. I have put in some hours reading the code and it is quite hard to follow because of what I  call voodoo magic code. This type of code causes and IDE to not be able to determine a data path through the code which makes it even harder to understand what is happening. In the hours I have spent reading over the code I have found several places similiar to the one you have pointed out above that can be changed to increase speed.

distutils also does not fix the `__cmp__` magic method and does not create a function  for `cmp()` which will cause a failure in the code. `__gt__`, `__lt__`, `__ge__`, `__le__`, `__eq__` and `__ne__` methods need to be added in place of `__cmp__` and the proper operators need to be used instead of using `cmp()`.

dosas: @kdschlosser That totally makes sense and I agree. 

@cmin764 I hope this gets merged and released soon

I would be happy to help with a PR to implement the changes you mentioned.

kdschlosser:
for those that need to put this PR into production This is he EXACT code that would be used when comtypes get installed with an older version of distutils (with a version of python 3.x that is older then 3.9). This is tested code and does work because it is what gets generated by distutils.

You CAN add this PR to specifically get used in one of 2 ways. The first way is by using requirements.txt. add the line below to your requirements.txt and it will install using the code in this PR

```
https://github.com/kdschlosser/comtypes/tarball/python_3
```

if you do not use requirements.txt you can optionally use setuptools.setup as well.

```
import setuptools

setuptools.setup(
    install_requires=['comtypes'],
    dependency_links=['https://github.com/kdschlosser/comtypes/tarball/python_3#egg=comtypes-1.1.10']
)
```

I cannot remember of any of the ways above requires the "wheel" package needs to be installed It could be only one of them or both. Someone would have to test to see.

PhilippSelenium:
@cfarrow @vasily-v-ryabov Could you review and release this?

kdschlosser:
I am going to create  a new branch on my fork of the repo that will have the changes made that are seen here and I will start to remove all of the Python 2 code  and also clean up the code so it is a little easier on the eyes.

I did also want to go through and make all of the structures that are declared so they are identical to what the Windows SDK is and add properties to get and set any field names that may not be the same. For backwards compatibility. I would also like to make sure that documentation is done with links to Microsoft's documentation. I remember when I first started using comtypes and how long it took me to figure out how in the world to use it.

I created a new issue https://github.com/enthought/comtypes/issues/256 that has a file attached to it. In that file is a whole mess of Windows data types that are not in comtypes or in ctypes.wintypes. It's a convenience  thing and it would be nice if comtypes also had something along these lines added to it.

One of the other things I have  thought about is how much information there is in the Windows SDK documentation and most of that information does pertain to comtypes and also ctypes. One of the things that I have thought would be super helpful is keeping the same file names as the header files. For example, ITypeInfo is located in typeinfo.py in comtypes but in the Windows SDK it is located in oaidl.h so when I look something up in the Windows SDK it would be really nice if if this was possible `from comtypes.oaidl import ITypeInfo`. I know that is kind of a small stupid thing but it would make it easier to use.


For those of you that use comtypes a lot here is another cookie for ya. 
This is no where near perfect and it does have errors in it so double checking against the  Windows SDK documentation needs to be done. It is not intended to be run. It is for reference purposes and also a time saver (copy the code).
  
https://github.com/kdschlosser/pyWinAPI

kdschlosser:
Here is that branch I mentioned above. You can do PR's against this and if someone starts merging PR's on this repo then we can do PR's of the changes made there.

https://github.com/kdschlosser/comtypes/tree/comtypes_update

cfarrow:
Thanks for the PR. I don't have a working environment for creating a build. @vasily-v-ryabov Do you have the bandwidth to make a release?

comtypes/__init__.py
cmin764: Yes, I can help you with getting it merged, then following a release.

vasily-v-ryabov:
@cfarrow I think I can make a release on the weekend. If @cmin764 wants to do that, please let me know.

vasily-v-ryabov:
I also plan to review important PR #241 first.

cmin764:
> I also plan to review important PR #241 first.

Waiting for you to finish with that, then I can rebase the release PR: https://github.com/enthought/comtypes/pull/258
What's the best approach for testing, releasing and publishing this package? (I used **build** and **twine** packages on my fork, from a Mac)

vasily-v-ryabov:
@cmin764 I'm not sure about Mac, I never tried such cross-OS release. Using twine has no alternatives.

kdschlosser:
Thanks guys for getting this PR done.

Make sure you put a notice about Python 2 no longer being supported. I am still working through and cleaning out other Python 2 relevant code and I will do a separate PR for that when it's done.

comtypes/server/inprocserver.py
```diff
@@ -119,7 +119,7 @@ def DllGetClassObject(rclsid, riid, ppv):
         if not cls:
             return CLASS_E_CLASSNOTAVAILABLE
 
-        result = ClassFactory(cls).IUnknown_QueryInterface(None, ctypes.pointer(iid), ctypes.c_void_p(ppv))
```
vasily-v-ryabov, line 122: This change is not clear to me in context of Py3 support.

kdschlosser: That change is from #234 that was merged.

mrzk:
Any update on when we will get the release?

vasily-v-ryabov:
This week.

PR #254, opened 2022-01-18
Python 3 setup support (setuptools downgrade)

cmin764:
Pins `setuptools<=57.5.0` in order to `pip install` with build-time Python 2 to 3 conversion support, otherwise you'll end up with the Python 2 flavor of the package under Python 3 envs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+.eggs/
+*.egg-info/
+build/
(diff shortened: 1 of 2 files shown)


cmin764:
Closing this in favor to: https://github.com/enthought/comtypes/pull/253

PR #259, opened 2022-01-30
Restore Python2.7 support without 2to3 fixer + timestamping fix

vasily-v-ryabov:
I want to keep Python2.7 support so far. It's easy without 2to3 fixer. Of course Py2.6 support is finally dropped. So we need to set it in setup.py, PyPI tags and CHANGES.txt.

Also I've rebased PR #241 to this PR keeping author of the commit.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,10 +3,11 @@ Comtypes CHANGELOG
 
 Release 1.1.11
 -------------
- * Fix Python 3.9+ compatibility. Thanks @kdschlosser.
+ * Fix setuptools>=60.0.0 compatibility. Thanks @kdschlosser.
  * Fix timestamping for typelib. Thanks @fusentasticus.
  * Fix 64bit inproc server.
- * Python 2.6 support is dropped.
+ * Drop Python 2.6 support.
+ * Fix IndexError for empty safearray. Thanks @BALOGHBence.
 
 Release 1.1.10
 -------------
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,9 +3,10 @@ Comtypes CHANGELOG
 
 Release 1.1.11
 -------------
- * Fix Python 3 compatibility. Thanks @kdschlosser.
- * Fix `tlib_mtime` in the absence of typelib.
+ * Fix Python 3.9+ compatibility. Thanks @kdschlosser.
+ * Fix timestamping for typelib. Thanks @fusentasticus.
  * Fix 64bit inproc server.
+ * Python 2.6 support is dropped.
 
 Release 1.1.10
 -------------
--- a/comtypes/GUID.py
+++ b/comtypes/GUID.py
@@ -8,6 +8,13 @@ else:
     def binary(obj):
         return buffer(obj)
 
+if sys.version_info >= (3, 0):
+    text_type = str
+    base_text_type = str
+else:
+    text_type = unicode
+    base_text_type = basestring
+
 BYTE = c_byte
 WORD = c_ushort
 DWORD = c_ulong
@@ -32,10 +39,10 @@ class GUID(Structure):
 
     def __init__(self, name=None):
         if name is not None:
-            _CLSIDFromString(str(name), byref(self))
+            _CLSIDFromString(text_type(name), byref(self))
 
     def __repr__(self):
-        return 'GUID("%s")' % str(self)
+        return 'GUID("%s")' % text_type(self)
 
     def __unicode__(self):
         p = c_wchar_p()
@@ -62,7 +69,7 @@ class GUID(Structure):
         return hash(binary(self))
 
     def copy(self):
-        return GUID(str(self))
+        return GUID(text_type(self))
 
     def from_progid(cls, progid):
         """Get guid from progid, ...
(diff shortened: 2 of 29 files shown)


vasily-v-ryabov:
@kdschlosser FYI

vasily-v-ryabov:
@cmin764 please test it with your cases. I've tried only `import comtypes` on Python 2.7.8 32-bit which works.

cmin764:
> @cmin764 please test it with your cases. I've tried only `import comtypes` on Python 2.7.8 32-bit which works.

We're using Python 3 only, so not sure if this is relevant anymore. Best case scenario for our testing part is to upgrade to `1.1.11` (after you fully release, tag & publish the package) and see if there is any regression with it. (as we're installing the package with Python 3 and we had Syntax Error issues because of the Py2 code not being converted during the build time due to the setuptools upgrade)

CHANGES.txt
```diff
@@ -3,9 +3,11 @@ Comtypes CHANGELOG
 
 Release 1.1.11
 -------------
- * Fix Python 3 compatibility. Thanks @kdschlosser.
- * Fix `tlib_mtime` in the absence of typelib.
+ * Fix setuptools>=60.0.0 compatibility. Thanks @kdschlosser.
```
cmin764, line 6: I think the version that introduced the [problem](https://github.com/robocorp/comtypes-fork/pull/1/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R21) was actually `57.5.0`.

comtypes/GUID.py
```diff
@@ -8,6 +8,13 @@ def binary(obj):
     def binary(obj):
         return buffer(obj)
 
+if sys.version_info >= (3, 0):
+    text_type = str
+    base_text_type = str
+else:
+    text_type = unicode
+    base_text_type = basestring
```
cmin764, line 16: What if we move all these definitions into a settings/constants module and then we just import them from there everywhere we use them? (so we don't repeat with this branching everywhere and in case we want to change a type, we just change it in one place (as it's imported everywhere else is used))

cmin764 (approved):
Good job on the PR, haven't tested it yet, but LGTM!

Although don't rely on us to test this version as we already got past our Syntax Error blocker with our fork and we'll be waiting for the wide community response on this new release before upgrading to the upcoming `1.1.11`.

dmwyatt:
FWIW, this cleared up my `Typelib different than module` timestamp errors on python 3.9.

kdschlosser:
I put in 2 more PR's that need to be looked at before you do another release.

comtypes/GUID.py
kdschlosser: That's a good idea

dmwyatt:
This adds a couple of SyntaxErrors.

You can't just test for python 3 with `sys.version_info` and then use the old `long` integer type because any usages of integers ending with `L` will raise a SyntaxError...even if you gate them behind an if/else checking for python3.

The two places a regex search for `\dL` I can find are:  

1. https://github.com/enthought/comtypes/blob/177aa55394264e57e7e5dbdadb15b0ae35d02c19/comtypes/test/test_variant.py#L117
2. https://github.com/enthought/comtypes/blob/177aa55394264e57e7e5dbdadb15b0ae35d02c19/comtypes/automation.py#L555

```python
In [1]: import comtypes.automation

  File ~\AppData\Local\pypoetry\Cache\virtualenvs\fhsvc-tqYogT0N-py3.9\lib\site-packages\IPython\core\interactiveshell.py:3251 in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  Input In [1] in <module>
    import comtypes.automation

  File ~\AppData\Local\pypoetry\Cache\virtualenvs\fhsvc-tqYogT0N-py3.9\lib\site-packages\comtypes\__init__.py:1216 in <module>
    class IPersist(IUnknown):

  File ~\AppData\Local\pypoetry\Cache\virtualenvs\fhsvc-tqYogT0N-py3.9\lib\site-packages\comtypes\__init__.py:1220 in IPersist
    COMMETHOD([], HRESULT, 'GetClassID',

  File ~\AppData\Local\pypoetry\Cache\virtualenvs\fhsvc-tqYogT0N-py3.9\lib\site-packages\comtypes\__init__.py:1138 in COMMETHOD
    from comtypes.automation import VARIANT

  File ~\AppData\Local\pypoetry\Cache\virtualenvs\fhsvc-tqYogT0N-py3.9\lib\site-packages\comtypes\automation.py:555
    v._.VT_I4 = 0x80020004L
                          ^
SyntaxError: invalid syntax


In [2]:
```

fusentasticus:
@dmwyatt  I raised the issue here https://github.com/enthought/comtypes/issues/262

kdschlosser:
The code could be changed to this tho


```python

try:
    long
except NameError:
    long = int


v._.VT_I4 = long(0x80020004)

```

kdschlosser:
Is the trailing L even needed in Python 2.7??

dmwyatt:
Maybe not?  

```python
Python 2.7.18 (v2.7.18:8d21aa21f2, Apr 20 2020, 13:25:05) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> 0x80020004L == 0x80020004
True
```

I can't imagine any reason to keep it, but that might say more about me then anything else...

kdschlosser:
I didn't think it needed to be there in python 2.7, I wasn't going to install Python 2.7 to find out  either 😁

comtypes/GUID.py
vasily-v-ryabov: That's good idea, but the time is very limited. If someone will take care of it, I can review and merge it as separate PR. Currently I'm going to release 1.1.11 without these cosmetics.

PR #260, opened 2022-01-31
Adds #232 back in

kdschlosser:
These code changes got removed because of a mistake I made when I submitted a PR.

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -267,9 +267,7 @@ class Generator(object):
             loaded_typelib = comtypes.typeinfo.LoadTypeLib(self.filename)
             full_filename = comtypes.tools.tlbparser.get_tlib_filename(loaded_typelib)
 
-            if full_filename is None:
-                tlib_mtime = 0
-            else:
+            if full_filename is not None:
                 # get DLL timestamp at the moment of wrapper generation
                 tlib_mtime = os.stat(full_filename).st_mtime
 
--- a/comtypes/server/inprocserver.py
+++ b/comtypes/server/inprocserver.py
@@ -119,7 +119,7 @@ def DllGetClassObject(rclsid, riid, ppv):
         if not cls:
             return CLASS_E_CLASSNOTAVAILABLE
 
-        result = ClassFactory(cls).IUnknown_QueryInterface(None, ctypes.pointer(iid), ppv)
+        result = ClassFactory(cls).IUnknown_QueryInterface(None, ctypes.pointer(iid), ctypes.c_void_p(ppv))
         _debug("DllGetClassObject() -> %s", result)
         return result
     except Exception:


vasily-v-ryabov:
The first changes is back in PR #259. Please check the second one.

kdschlosser:
#232 has not been added back in. 

I will fix the merge conflicts with this PR and then all should be good to go.

comtypes/tools/codegenerator.py
```diff
@@ -276,12 +276,17 @@ def generate_code(self, items, filename):
         for line in wrapper.wrap(text):
             print(line, file=self.output)
 
-        if os.path.exists(filename):
```
vasily-v-ryabov, line 279: I thought it is part of the fix from @fusentasticus, isn't it?

vasily-v-ryabov: See PR #241

kdschlosser: What happened was when I did the initial PR for correcting the python3 problems is I used Python 3.8 and I installed comtypes using pypi.  This allowed the code conversion from python 2 to 3 to run. I then used the  generated  code and created a PR.

Me being brain dead at that time didn't think about any PR's that had been merged since 1.1.10 was released. The  PR I had submitted  got merged and that effectively undid 2 PR's that had gotten merged. Once of them you fixed and the other  that still needs to be fixed is this PR.

All I did was I looked at the code in the comtypes repo just prior to the Python3 fix getting merged and just after it getting merged and compared the differences. I looked for any changes that didn't make any sense for a Python 2 to 3 code conversion  and then looked to see if there was a PR that aligned with it. The result of that was this PR.

All this PR does is return the section of code that shouldn't have been changed back to what it was before the Python3 fix PR got merged, nothing more. If there are additional changes you are wanting to do above and beyond that I think should be done in a separate PR instead of this one. You are the boss and whatever you would like to do is what will get done.

I want to make sure that you understand fully what the purpose of this PR is before moving forward.

vasily-v-ryabov: Don't worry, I also rely on contributors testing because I'm usually overloaded. I'd prefer to close this PR for now. Let's see how new release will run on a wide audience. If this place will become problematic again, we can get back to it. Hope your own scenarios are not broken without this PR. This is what is good to know. Thanks for your efforts! Really appreciate it!

kdschlosser:
Everything is good on my end . I don't need to have this code change for my stuff to work. I don't use the typelib code generator at all. I wrote a script that reads the Windows SDK header and IDL files and converts it to python. While it is not perfect it does reduce a whole lot of typing. For a 10,000 line  header file I might have to spend 30 minutes scrolling over it  to correct any errors the script may have had. I have about a million lines of the windows SDK ported to Python. It's good to have as a reference and also using to just copy and paste instead of having to key  out what I may need.

It's located here.

https://github.com/kdschlosser/pyWinAPI

The directory and file structure is identical to the Windows SDK.

PR #261, opened 2022-01-31
Removes build_py_2to3 and adds import of print_function

kdschlosser:
`__future__.print_function` needs to be imported in order to have the `print()` statements work properly on Python 2.7.

Because the code syntax is Python 2.7 compliant there is no longer a need to use build_py_2to3 to do a code conversion. So I have removed this from setup.py

### DIFF
--- a/comtypes/client/_events.py
+++ b/comtypes/client/_events.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import ctypes
 import traceback
 import comtypes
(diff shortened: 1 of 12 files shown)


vasily-v-ryabov:
Nice catch! Please rebase it from master and resolve one simple conflict.

kdschlosser:
you should be good to go.

kdschlosser:
Thank you for the "Nice catch" comment it's appreciated and also merging this.

I can understand why the want to keep support for Python 2.7. I have a project that uses Python 2.7 and a conversion to Python3 is going to be a HUGE amount of work considering the project is 250,000+ lines of code. That project does use comtypes 😉. You have probably heard of of the project, it's called EventGhost. EventGhost has been around some 15 years and has been using comtypes since the beginning I believe, or really close to it anyhow.

PR #272, opened 2022-02-06
Update _comobject.py

piyush8358:
try to fix some invalid Syntax


vasily-v-ryabov:
Hi @piyush8358 did you test current master branch? It's not released yet, but it will be released this week. I think this PR reverts some useful changes and it should break Py3 compatibility.

piyush8358:
ok I try to resolve  that  shouldn't break compatibility

PR #273, opened 2022-02-08
Update GUID.py

piyush8358:
Try to fix some error  , test are worked proper now

### DIFF
--- a/comtypes/GUID.py
+++ b/comtypes/GUID.py
@@ -54,7 +54,7 @@ class GUID(Structure):
 
     def __cmp__(self, other):
         if isinstance(other, GUID):
-            return cmp(binary(self), binary(other))
+            return (binary(self), binary(other))
         return -1
 
     def __bool__(self):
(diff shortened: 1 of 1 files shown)


comtypes/GUID.py
```diff
@@ -54,7 +54,7 @@ def __unicode__(self):
 
     def __cmp__(self, other):
         if isinstance(other, GUID):
-            return cmp(binary(self), binary(other))
+            return (binary(self), binary(other))
```
vasily-v-ryabov, line 57: It looks strange to me because `__cmp__` method should return "a negative value if self < other , positive if self > other , and zero if they were equal" (c). So why does it return a 2-tuple now???

piyush8358:
it shows the problem that CMP is not defined if we add brackets like   
def __cmp__(self, other):
        if isinstance(other, GUID):
            return cmp()(binary(self), binary(other))
        return -1
        it works properly

piyush8358 (commented):
def __cmp__(self, other):
        if isinstance(other, GUID):
            return cmp()(binary(self), binary(other))
        return -1

vasily-v-ryabov:
This is wrong anyway. Method `__lt__` is what we need.

piyush8358:
Ok, Thank you for the update. I try to improve
.

PR #274, opened 2022-02-11
Loading registered type libraries using a .NET name

kdschlosser:
Not all .NET names can be used because there isn't typelib information available for all of them.

Here are a few examples

If you have an Nvidia video card
`comtypes.typelib.DisplayServer.Config`

Windows Media Player
`comtypes.typelib.MediaPlayer.MediaPlayer`

### DIFF
--- a/comtypes/typelib_loader.py
+++ b/comtypes/typelib_loader.py
@@ -1,0 +1,255 @@
+import sys
+from comtypes.client import GetModule
+import winreg
+
+# this makes it easy to load registered type libraries. It uses the .NET
+# names that are stored in the registry to function. Not all .NET names
+# can be used because there isn't typelib information available for all
+# of them.
+# Here are a few examples
+#
+# if you have an Nvidia video card
+# comtypes.typelib.DisplayServer.Config
+#
+# Windows Media Player
+# comtypes.typelib.MediaPlayer.MediaPlayer
+
+
+def _get_reg_value(path, key, wow6432=False):
+    d = _read_reg_values(path, wow6432)
+    if key in d:
+        return d[key]
+
+    return ''
+
+
+def _read_reg_keys(key, wow6432=False):
+    if isinstance(key, tuple):
+        root = key[0]
+        key = key[1]
+    else:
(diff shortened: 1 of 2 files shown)


comtypes/typelib_loader.py
```diff
@@ -0,0 +1,255 @@
+import sys
+from comtypes.client import GetModule
+import winreg
```
vasily-v-ryabov, line 3: Need to wrap it for Py2.7 compatibility:
```python
if sys.version_info >= (3, 0):
    import winreg
else:
    import _winreg as winreg
```

comtypes/typelib_loader.py
```diff
@@ -0,0 +1,255 @@
...
+        self.__loader__ = mod.__loader__
+        self.__name__ = mod.__name__
+        self.__package__ = mod.__package__
+        # self.__path__ = mod.__path__
+        self.__spec__ = mod.__spec__
+        self.__original_module__ = mod
+
+        sys.modules[mod.__name__] = self
+
+    def __getattr__(self, item):
```
vasily-v-ryabov, line 108: `__getattribute__` is a recommended method for new-style classes. It should call `object.__getattribute__(self, item)` only when necessary so it eliminates infinite recursion.

comtypes/typelib_loader.py
```diff
@@ -0,0 +1,255 @@
...
+
+class _TypeLibLoader(object):
+
+    def __init__(self, keys, name=None):
+        self.__loaded = {}
+        self.__names = {}
+        self.__name = name
+        self.__keys = keys
+
+    def __getattr__(self, item):
```
vasily-v-ryabov, line 129: The same about `__getattribute__`

vasily-v-ryabov (commented):
Not sure if import hook is needed here. I've left just general comments.

comtypes/typelib_loader.py
vasily-v-ryabov: Well, OK. Maybe it's enough to have only `__getattr__`.

PR #275, opened 2022-02-11
Fixes tools.codegenerator.Generator.generate_code

kdschlosser:
I ran across an issue where I had exported a typelib resource from a dll file and the resulting tlb file when fed into the code generator would fail because of a version mismatch. This was happening because the file stored in the typelib was different from the file that was being used to generate the code.

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -248,8 +248,28 @@ class Generator(object):
                 assert os.path.isfile(p)
 
     def generate_code(self, items, filename):
+        tlib_mtime = None
+
+        if filename is not None:
+            # get full path to DLL first (os.stat can't work with relative DLL paths properly)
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov:
@fusentasticus @dmwyatt could you please test these changes on your cases? If it works, I will merge it.

dmwyatt:
It works for me.

kdschlosser:
There is no rush on this. It  was something I stumbled upon when writing #274

There is also another big that I will be making a PR for but instead of just fixing that single bug I am going to eliminate any other possible bugs. It has to deal with the code generator as well but specifically the adding of non ctypes objects to a list of known ctypes objects. If that makes any sense to ya then you will kind of understand what I am getting at.

vasily-v-ryabov:
Looks like this PR should be rebased to resolve conflicts.

kdschlosser:
I messed up the rebase and I created a new PS to keep the commit history clean. The new PR is #281

PR #278, opened 2022-02-13
Fixes tests

kdschlosser:
***WIP***

This is the start of getting the tests up and running again. Quite a few of them are running properly and they are passing. I have an error in the test_jscript.js file I am not sure where that's coming from. Maybe someone else could shed some light on that bit of the code for me.

 At this point the above error is not my major concern. The tests don't finish and I get an application crash 0xC0000005 (Access Violation). I have to look into what is going on with this.
 
 I did fix compiling the typelib in test_server.py.
 
 Over the last 2 years or so I have been working on a script that creates an MSVC build environment for compiling Python extension modules and C libraries. The one that both distutils and setuptools creates is riddled with problems and neither of them work on all MSVC versions. The script I wrote DOES NOT use the vcvars*.bat files. to create the build environment. Those batch files are buggy and don't always setup the build environment properly.
 
It's kind of circular because the script uses comtypes to create the build environment. While it can do it without as a backup it is much faster with it. So even if comtypes  is broken in a manner that does not allow that portion of the script to run the server test type lib will still get compiled properly.

The MSVC script works for Visual Studio 2008 up to the most current Visual Studio 2022, Visual C++ 2008 to current Visual C++ 2022 All versions of  Build Tools. It supports all .NET and NETFX versions and all Windows SDK's from 6.0a to the latest Windows 11 SDK.

It should work without issue for many years.

I forgot, it detects the Visual C++ Compiler for Python 2.7 as well.

Any help with getting this running would be awesome!


bennyrowland:
@kdschlosser this looks very cool to me - in my `comtypes` server projects I have been using [`scikit-build`](https://scikit-build.readthedocs.io/en/latest/) to set-up my VS environment and then CMake to run the MIDL compiler and insert the .TLB file into my package. This works well for me, but it does take a long time (especially combined with PEP517 isolated builds) and I haven't tested it over a wide range of environments and versions. Your MSVC script looks like it could be very useful for that kind of problem without all the trappings of `scikit-build` (e.g. for testing rather than building as here). Is there any reason why you haven't set it up as its own package on PyPI? It feels a bit big to stick into the `comtypes` test folder and I am sure there are lots of other set-ups that would benefit from it in the more general case.

The circular dependency issue with `comtypes` is a bit of a pain, but it could be addressed by moving the `comtypes` tests out of the package (to the top level of the repo). Then the tests are an independent compontent which have the combined dependencies of `comtypes` and `msvc`, rather than being part of a package that requires another package that depends on itself. Personally, I prefer the tests outside the package anyway, not only do they bulk out the installable wheel unnecessarily but probably have various additional dependencies that needn't be installed by a user.

I am definitely interested in working on getting the `comtypes` server side of things working better and particularly getting tests working, perhaps we can join forces on this?

PR #279, opened 2022-02-13
Cleans up generated code output.

kdschlosser:
This is a pretty involved PR in what it does. There is a decent amount of code changes but nothing to crazy. Here are the things the PR does

removes duplicate imports
removes unused imports
puts declarations after imports instead of being mixed in with the imports
formats most code that is over 80 characters on a single line
2 blank lines before and after classes
puts class docstrings as the first line in the class  this way it can actually get set to `__doc__`
fixes lack of triple double quotes not being used around class docstrings

So in a nutshell it makes most of the generated code PEP8 compliant and it makes the generated code a whole hell of a lot easier on the :eyes:

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -167,12 +167,14 @@ class Generator(object):
         self._externals = {}
         self.output = ofi
         self.stream = io.StringIO()
-        self.imports = io.StringIO()
+        self.imports = {}
+        self.declarations = io.StringIO()
 ##        self.stream = self.imports = self.output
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov:
pywinauto tests with this PR branch: https://ci.appveyor.com/project/pywinauto/pywinauto-8b7t7/builds/42558727

One Python 3.6 line fails expectedly. The rest should pass. Hopefully.

vasily-v-ryabov (approved):
I can say it's good enough after quick look. If pywinauto tests are not broken, I will merge it.

Issue #293, opened 2022-05-14
AttributeError: module 'comtypes.gen._...' has no attribute 'IEnumVARIANT'


junkmd:
When I run some code with `comtypes` version 1.1.12, raises `NameError`(#284).

Even if I fixed this error(#287), other error has raised in some modules.

```py
Python 3.10.0 (tags/v3.10.0:b494f59, Oct  4 2021, 19:00:18) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import comtypes
>>> comtypes.__version__
'1.1.12'
>>> from comtypes.client import GetModule
>>> GetModule(('{D3295D87-D604-11D4-A704-00C04FA137E4}', 8, 0))  # Microsoft_JScript
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\...\comtypes\comtypes\client\_generate.py", line 159, in GetModule
    mod = _CreateWrapper(tlib, pathname)
  File "C:\Users\...\comtypes\comtypes\client\_generate.py", line 227, in _CreateWrapper
    generate_module(tlib, ofi, pathname)
  File "C:\Users\...\comtypes\comtypes\tools\tlbparser.py", line 767, in generate_module
    gen.generate_code(list(items.values()), filename=pathname)
  File "C:\Users\...\comtypes\comtypes\tools\codegenerator.py", line 290, in generate_code
    self.generate_all(items)
  File "C:\Users\...\comtypes\comtypes\tools\codegenerator.py", line 206, in generate_all
    self.generate(item)
  File "C:\Users\...\comtypes\comtypes\tools\codegenerator.py", line 202, in generate
    mth(item)
  File "C:\Users\...\comtypes\comtypes\tools\codegenerator.py", line 856, in CoClass
    self.generate(itf.get_head())
  File "C:\Users\...\comtypes\comtypes\tools\codegenerator.py", line 202, in generate
    mth(item)
  File "C:\Users\...\comtypes\comtypes\tools\codegenerator.py", line 799, in External
    comtypes.client.GetModule(ext.tlib)
  File "C:\Users\...\comtypes\comtypes\client\_generate.py", line 159, in GetModule
    mod = _CreateWrapper(tlib, pathname)
  File "C:\Users\...\comtypes\comtypes\client\_generate.py", line 242, in _CreateWrapper
    mod = _my_import(fullname)
[...]

PR #287, opened 2022-03-17
Fix NameError in generated modules

junkmd:
https://github.com/enthought/comtypes/issues/284#issuecomment-1071665989

Fix broken imports of generated modules.

Issue #284, opened 2022-03-05
NameError error


zboolk:
**I have the code like this:**
```python
import comtypes 
from comtypes import * 
from comtypes.client import * 

comtypes.client.GetModule('UIAutomationCore.dll') 
from comtypes.gen.UIAutomationClient import * 

uia = CreateObject(CUIAutomation._reg_clsid_, interface=IUIAutomation) 

from pywinauto.application import Application

app = Application(backend="uia").start("notepad.exe")
app.UntitledNotepad.type_keys("%FX")
```

**when I run it with Python 3.8.7（comtypes version 1.1.12） , it appears to show me the error, I don't know how to fix it:**
```python
C:\stock\program>python robot.py
Traceback (most recent call last):
  File "robot.py", line 5, in <module>
    comtypes.client.GetModule('UIAutomationCore.dll')
  File "C:\Users\zbool\AppData\Roaming\Python\Python38\site-packages\comtypes\client\_generate.py", line 159, in GetModule
    mod = _CreateWrapper(tlib, pathname)
  File "C:\Users\zbool\AppData\Roaming\Python\Python38\site-packages\comtypes\client\_generate.py", line 227, in _CreateWrapper
    generate_module(tlib, ofi, pathname)
  File "C:\Users\zbool\AppData\Roaming\Python\Python38\site-packages\comtypes\tools\tlbparser.py", line 767, in generate_module
    gen.generate_code(list(items.values()), filename=pathname)
  File "C:\Users\zbool\AppData\Roaming\Python\Python38\site-packages\comtypes\tools\codegenerator.py", line 290, in generate_code
    self.generate_all(items)
  File "C:\Users\zbool\AppData\Roaming\Python\Python38\site-packages\comtypes\tools\codegenerator.py", line 206, in generate_all
    self.generate(item)
  File "C:\Users\zbool\AppData\Roaming\Python\Python38\site-packages\comtypes\tools\codegenerator.py", line 202, in generate
    mth(item)
  File "C:\Users\zbool\AppData\Roaming\Python\Python38\site-packages\comtypes\tools\codegenerator.py", line 882, in ComInterface
    self.generate(itf.get_head())
  File "C:\Users\zbool\AppData\Roaming\Python\Python38\site-packages\comtypes\tools\codegenerator.py", line 202, in generate
    mth(item)
[...]

PR #294, opened 2022-05-29
fix `tools.codegenerator` for `NameError` and `AttributeError` in generated modules.

junkmd:
#293 

Fix `NameError` and `AttributeError` in generated modules.

## Relationship with #287
I PRed #287 to fix lack of `from ctypes import *` line in generated modules(#284).
But another problem in namespaces of generated modules(#293).
This PR fixes both problems.

### Add [`ImportedNamespaces`](https://github.com/enthought/comtypes/commit/362fdd75#diff-af6c0d9585b5544b699e1be4e5a25f10797e989b9a983c2d91928f79c59c8a34R1169-R1236) and [`DeclaredNamespaces`](https://github.com/enthought/comtypes/commit/362fdd75#diff-af6c0d9585b5544b699e1be4e5a25f10797e989b9a983c2d91928f79c59c8a34R1239-R1267)

They will be [`Generator` attribute](https://github.com/enthought/comtypes/commit/362fdd75#diff-af6c0d9585b5544b699e1be4e5a25f10797e989b9a983c2d91928f79c59c8a34R173-R174) instead of `dict` or `io.StringIO`.
Formerly, to avoid duplicating import and alias-declaring lines, `Generator._Foo_defined` attributes are necessary.
With this change, `ImportedNamespaces` and `DeclaredNamespaces` prevent duplication of generated modules' lines instead of `Generator` itself.

### `self.output` is used only when [`Generator.generate_code`](https://github.com/enthought/comtypes/commit/362fdd75#diff-af6c0d9585b5544b699e1be4e5a25f10797e989b9a983c2d91928f79c59c8a34R292-R309) was called

Formerly, many I/O(`self.output`, `self.declarations`, `self.stream`) and writing/printing to them are messy and will be a root of weird bugs.
With this change, `self.stream` is the I/O that will be mainly used and `self.output` is used only in `Generator.generate_code`.

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -9,6 +9,7 @@ else:
     import cStringIO as io
 import keyword
 import ctypes
+import textwrap
 
 from comtypes.tools import typedesc
 import comtypes
@@ -169,9 +170,8 @@ class Generator(object):
         self._externals = {}
         self.output = ofi
         self.stream = io.StringIO()
-        self.imports = {}
-        self.declarations = io.StringIO()
-##        self.stream = self.imports = self.output
+        self.imports = ImportedNamespaces()
+        self.declarations = DeclaredNameSpaces()
         self.known_symbols = known_symbols or {}
 
         self.done = set() # type descriptions that have been generated
@@ -186,9 +186,7 @@ class Generator(object):
         else:
             name = getattr(item, "name", None)
         if name in self.known_symbols:
-            mod = self.known_symbols[name]
-            if name not in self.imports:
-                self.imports[name] = mod
+            self.imports.add(name, symbols=self.known_symbols)
 
             self.done.add(item)
             if isinstance(item, typedesc.Structure):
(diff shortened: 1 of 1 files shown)


comtypes/tools/codegenerator.py
```diff
@@ -1274,6 +1165,108 @@ def make_DispProperty(self, prop):
...
+
+class ImportedNamespaces(object):
+    def __init__(self):
+        if sys.version_info >= (3, 7):
+            self.data = {}
+        else:
+            from collections import OrderedDict
+            self.data = OrderedDict()
+
+    def add(self, *names, symbols=None):
```
vasily-v-ryabov, line 1177: This syntax (`*names` and further keyword args) is not Py2.7 compatible. I still want to keep it for pywinauto project.

junkmd: I made this py27-compatible.

P.S.
`*names, symbols=None` syntax was not failed in AppVeyor CI.
Perhaps, some tests are broken.

IMO release of 1.1.13 should wait until tests were fixed.

junkmd:
I fixed class name case(`DeclaredNameSpaces` -> `DeclaredNamespaces`, adjust to `ImportedNamespaces`).

comtypes/tools/codegenerator.py
vasily-v-ryabov: `comtypes` AppVeyor tests are very limited. There are only 2 install tests. The vast majority of functional tests are not enabled in AppVeyor for now. I've detected this Py2.7 issue on pywinauto tests (on pywinauto's AppVeyior). If they pass, I'm probably going to merge this PR. Thanks for your efforts!

vasily-v-ryabov (approved):

PR #297, opened 2022-06-05
Please see re-submitted PR #299

junkmd:
`comtypes.named_property` and `comtypes.bound_named_property` are [descriptor](https://docs.python.org/3/library/functions.html#property), behave like builtin [`property`](https://docs.python.org/3/library/functions.html#property).

They have same names args and attributes, but there are differences in their roles.

For example, `getter` is
- an instance attribute of `named_property`.
- a method of builtin `property`.

It is very complicated, so this PR changes `(bound_)named_property` to match the `property`.

## Changes
- `getter` -> `fget`
- `setter` -> `fset`
- `im_inst` -> `instance`
- `im_class` -> `owner`

## FYI
- In python 3.10, `inspect`ed result of builtin `property` is below.
```py
>>> import inspect
>>> inspect.signature(property)
<Signature (fget=None, fset=None, fdel=None, doc=None)>
>>> inspect.signature(property.__get__)
<Signature (self, instance, owner, /)>
>>> inspect.signature(property.__set__)
<Signature (self, instance, value, /)>
>>> property.getter
<method 'getter' of 'property' objects>
>>> property.setter
<method 'setter' of 'property' objects>
```

- https://docs.python.org/3/howto/descriptor.html#properties

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,9 +4,9 @@ max_jobs: 3
 
 init:
   - git config --global core.autocrlf input
-  
+
 shallow_clone: true
- 
+
 environment:
    matrix:
      - py: Python27
@@ -28,3 +28,4 @@ test_script:
    - C:\%py%\python.exe setup.py install
    - C:\%py%\Scripts\pip.exe uninstall comtypes -y
    - C:\%py%\python.exe test_pip_install.py
+   - C:\%py%\python.exe -m unittest discover -v -s ./comtypes/test -t comtypes\test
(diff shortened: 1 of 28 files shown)


junkmd:
I rebased this branch for doing tests on AppVeyor CI.

vasily-v-ryabov:
I'd prefer to make rebase without merge commit. The changed files look strange for this PR (very similar to previous PR which was already merged). Maybe some GitHub glitch.

junkmd:
@vasily-v-ryabov 

OK, I'll close this PR and submit new one instead of resolving the glitch.

junkmd:
Please see #299

PR #300, opened 2022-06-11
Refactor `comtypes.client._generate.GetModule`

junkmd:
`client.GetModule` is one of the most used functions in `comtypes`, but the implemented code is quite a lot of lines.

Therefore, I have refactored it by splitting functions according to their roles.

And from the point of view of #114 by @vijairaj,
> the code is buffered in to StringIO before making it to the file and if there's an interruption in between, it would not make it to the file and result in an incomplete file
https://github.com/enthought/comtypes/issues/114#issuecomment-284348343

so I fixed the timing of when the generated code is written to the file.

### DIFF
--- a/comtypes/client/_generate.py
+++ b/comtypes/client/_generate.py
@@ -19,7 +19,9 @@ else:
     import _winreg as winreg
     import cStringIO as io
 
+
 PATH = os.environ["PATH"].split(os.pathsep)
+
 
 def _my_import(fullname):
     # helper function to import dotted modules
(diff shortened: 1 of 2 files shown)


junkmd:
@vasily-v-ryabov 

Your AppVeyor tests have been successful (except for `reportlab` installation failed in Python 3.6).
Do you have any concerns with this PR?

vasily-v-ryabov:
All is fine. I was busy/overloaded last 2 weeks. My students were preparing their master's degree presentations about their work on pywinauto. :)

junkmd:
@vasily-v-ryabov 
Ok.
Thanks for taking the time out of your busy schedule to review this.

Issue #301, opened 2022-06-13
Enhance `client.Constants`


junkmd:
I wish enhancing `client.Constants`.

- enable instanciate from typelib path or progid-like string,
```py
>>> from comtypes.client import Constants
>>> c = Constants('UIAutomationCore.dll') 
>>> c.StyleId_Title  
70010
```

- enable access to enumerations
```py
>>> from comtypes.client import Constants
>>> c = Constants('UIAutomationCore.dll') 
>>> c.LiveSetting.Assertive
2
```

junkmd:
## other additional features (I wish)
- defines [`__slots__`](https://docs.python.org/3/reference/datamodel.html?highlight=slots#slots) to
  - disable be assigned new variables
  - prevents the automatic creation of `__dict__` and `__weakref__` for each instance
- adds frozen dict that can access to members by using dot to
  - represent enumerations
    - instead of [`enum.IntEnum`](https://docs.python.org/3/library/enum.html) that is New in version 3.4
  - prevent overwriting `Constants` instance attributes' members

dot-accessible frozen dict implementation maybe
```py
class _frozen_attr_dict(dict):
    __slots__ = ()

    def __getattr__(self, name):
        if name not in self:
            raise AttributeError
        return self[name]

    def __setitem__(self, key, value):
        raise TypeError

    def __delitem__(self, name):
        raise TypeError

    def __ior__(self, other):
        # `dict |= other` is New in version 3.9,
        # but this class does not support it.
        raise TypeError

    def clear(self):
        raise TypeError

    def pop(self, key, default=None):
        raise TypeError

    def popitem(self, last=True):
        raise TypeError

    def setdefault(self, key, default=None):
        raise TypeError
```

PR #305, opened 2022-06-18
add TestCase for `client.Constants`

junkmd:
I submitted #301 for enhancements of `client.Constants`.

But I noticed that there is no test to check the current `Constants` behavior.

With this change, the TestCase for `Constants` is added to `test_client.py`


junkmd:
@vasily-v-ryabov 
Thank you for resolving conflictions.

PR #315, opened 2022-06-26
No patching or modifying `typedesc.Obj.name` in `tools.codegenerator`

junkmd:
Formerly,
- There is destructive patching or modifying attribute by non-`self` object. It will may be root of bug in future.
- `codegenerator.Generator.type_name` is not only returning type name but including `self.declarations.add(...)`.

With this change, in `tools.codegenerator.Generator`,
- no patching or modifying `typedesc.Obj.name`.
- using `self.type_name(obj)` instead of `obj.name`.
- dead `generate=True` arg was removed from `type_name` method.
- `self.type_name(obj)` returns name string only. Defining alias process was moved to another method.

And I removed workaround for `set` which is not needed now.


junkmd:
I merged master branch updated by #322.
I confirmed AppVeyor CI was not failed.

vasily-v-ryabov (approved):

PR #298, opened 2022-06-08
Make tests runnable on AppVeyor CI and environment-specific tests be skipped

junkmd:
related to #267

This PR has taken over from #271 by @dmwyatt.
- participants are @vasily-v-ryabov, @kdschlosser

With this change, tests can run on AppVeyor CI(and perhaps on developer machines).

The goal of this PR as is @dmwyatt's comment,
> It seems like it'd be better to have a test suite that runs but doesn't cover everything than a test suite that doesn't run but hypothetically covers more.
https://github.com/enthought/comtypes/issues/267#issuecomment-1029305670

## Overview of changes
links to issue/PR comments.

### Skips environment-specific tests
- [failed with `OSError: ... Invalid class string`](https://github.com/enthought/comtypes/issues/267#issue-1122368686)
- [failed with assertion in `test_leaks_x`](https://github.com/enthought/comtypes/issues/267#issue-1122368686)
- [failed with `ResourceDenied`](https://github.com/enthought/comtypes/issues/267#issue-1122368686)
- [depends upon IE and MS Office](https://github.com/enthought/comtypes/pull/271#issue-1125294924)
- [depends upon `comtypes/test/TestComServer.tlb` or `comtypes/test/TestDispServer.tlb`](https://github.com/enthought/comtypes/pull/271#issue-1125294924)
- [causes the whole test suite to fail for some reason](https://github.com/enthought/comtypes/pull/271#issue-1125294924)

### Fixes failed or warned tests
- [adds import conditional branch](https://github.com/enthought/comtypes/pull/295#issuecomment-1148740706)
- [replaces to non-environment-specific dll and asserts attributes' value instead of object](https://github.com/enthought/comtypes/pull/295#issuecomment-1148737128)
- [prevents `ResourceWarning`](https://github.com/enthought/comtypes/pull/295#issuecomment-1148720256)

Issue #212, opened 2020-06-22
Argument lifetime issue when calling COM/ATL methods with array of strings (and probably others) as of at least Python 3.8.3 (which worked with Python 3.7.x)


whannes:
Consider the following, very simple COM/ATL project made with MSVC 2017 (15.9.24, Windows 10 Pro 2004): [comProjectWithArray-v2.zip](https://github.com/enthought/comtypes/files/4845419/comProjectWithArray-v2.zip). It does not contain any functionality and is used to reproduce the empty array as described below only.

Registration is disabled in the project settings here, so to test it you have to register
`ATLTestServer.exe /RegServer`
`regsvr32 ATLTestServerPS.dll`

After successful registration, it can be tested with the following Python-Snippet:
`from comtypes.client import CreateObject`
`obj = CreateObject("ATLTestServer.ATLTestObject")`
`obj.concatenateStrings(["New", " ", "York"])`

When using comtypes 1.1.7 (6fb440372bb03e3acece00b35cac87f2eddf7d69) with Python 3.7.7 x64 the call from the Python snipped gets properly routed to the C++ method:
![concatCallPython37](https://user-images.githubusercontent.com/41183585/85300577-05ea4a00-b4a7-11ea-9f0d-2ca9475e1f59.png)

However, when using the same comtypes version with Python 3.8.3 x64, the array in C++ contains only nullptr:
![concatCallPython38](https://user-images.githubusercontent.com/41183585/85300653-1ef2fb00-b4a7-11ea-9cd9-7a9f3ebc7e1d.png)

Is this a known issue?

whannes:
Update: It seems it is not related to the cache files generated. Although they are not exactly the same on my machine with regard to the order of the definitions, using the ones created with Python 3.7 in Python 3.8 and vice versa does not changes the behavior: When calling the server with Python 3.8, the array of variants is empty while with Python 3.7 everything is fine.

karpierz:
Hi All
Any news?. IMHO it is very important bug/things which in practice makes comtypes impossible to use in Py >=3.8
Thanks,
Adam

vasily-v-ryabov:
@karpierz I see the importance of the issue. I will try to find time to work on it in June. Since pywinauto is not affected, it's P2 for me. Thanks for understanding.

karpierz:
Vasilij
I tried to investigate this issue in sources but.. without any succes/seems for me too difficult (I am not familiar with comtypes source for now;).
[...]

PR #316, opened 2022-06-26
Update `test_excel`

junkmd:
In #298, environment-specific-tests are skipped.

But there are `def setUpModule;raise SkipTest(...` in some modules.
In this way, skipped testcase names are not displayed even if using `-v` option.

With this change, in `test_excel`, skipped testcase name and reasons are displayed in console.

for example:
```
# with non-excel-installed environment
test (test_excel.Test_EarlyBind) ... skipped 'This depends on Excel.'
test (test_excel.Test_LateBind) ... skipped 'This depends on Excel.'
```

```
# with excel-installed environment
test (test_excel.Test_EarlyBind) ... skipped 'There is difference of `Range.Value` behavior between Python >= 3.8.x and Python <= 3.7.x.'
test (test_excel.Test_LateBind) ... ok
```

See https://github.com/enthought/comtypes/issues/212#issuecomment-1163176126 for details of `@unittest.skip("There is difference of  ...`

### DIFF
--- a/comtypes/test/test_excel.py
+++ b/comtypes/test/test_excel.py
@@ -24,7 +24,7 @@ except (ImportError, OSError):
     IMPORT_FAILED = True
 
 
-class Base(object):
+class BaseBindTest(object):
     # `dynamic = True/False` must be defined in subclasses!
 
     def setUp(self):
@@ -117,12 +117,12 @@ class Base(object):
 @unittest.skipIf(IMPORT_FAILED, "This depends on Excel.")
 @unittest.skip("There is difference of `Range.Value` behavior "
     "between Python >= 3.8.x and Python <= 3.7.x.")
-class Test_EarlyBind(Base, unittest.TestCase):
+class Test_EarlyBind(BaseBindTest, unittest.TestCase):
     dynamic = False
 
 
 @unittest.skipIf(IMPORT_FAILED, "This depends on Excel.")
-class Test_LateBind(Base, unittest.TestCase):
+class Test_LateBind(BaseBindTest, unittest.TestCase):
     dynamic = True
 
 
--- a/comtypes/test/test_excel.py
+++ b/comtypes/test/test_excel.py
@@ -24,10 +24,7 @@ except (ImportError, OSError):
     IMPORT_FAILED = True
 
 
-@unittest.skipIf(IMPORT_FAILED, "This depends on Excel.")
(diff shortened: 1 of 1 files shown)


comtypes/test/test_excel.py
```diff
@@ -4,32 +4,43 @@
...
+    from comtypes.gen.Excel import xlRangeValueDefault
+    IMPORT_FAILED = False
+except (ImportError, OSError):
+    IMPORT_FAILED = True
+
+
+@unittest.skipIf(IMPORT_FAILED, "This depends on Excel.")
+@unittest.skip("There is difference of `Range.Value` behavior "
+    "between Python >= 3.8.x and Python <= 3.7.x.")
+class BaseTest(unittest.TestCase):
```
vasily-v-ryabov, line 30: I would suggest to not inherit `BaseTest` from `unittest.TestCase` and do multiple inheritance in derived classes so they are inherited both from `BaseTest` and from `unittest.TestCase`. Declaring some class and then deleting it looks strange.

Another option is naming `test` method as non-test in base class and call it in derived class's test method which is a one liner or maybe an alias would work (didn't check it).

comtypes/test/test_excel.py
```diff
@@ -4,76 +4,76 @@
...
+    from comtypes.gen.Excel import xlRangeValueDefault
+    IMPORT_FAILED = False
+except (ImportError, OSError):
+    IMPORT_FAILED = True
 
-def setUpModule():
-    raise unittest.SkipTest("External test dependencies like this seem bad.  Find a different "
-                            "built-in win32 API to use.")
 
+class Base(object):
```
vasily-v-ryabov, line 27: I think `BaseBindTest` is better name to explain its content. `Base` sounds too general.

junkmd: OK, I fixed it.

junkmd: And I removed `@unittest.skip("There is difference of ...` from `Test_LateBind`.

`CreateObject(..., dynamic=True)` and `xl.Range["A1", "C1"].Value[()] = (10,"20",31.4)` does not set `(None, None, 10.0)` to `A1:C1` even if Python >= 3.8.x(see https://github.com/enthought/comtypes/issues/212#issuecomment-1163176126).

vasily-v-ryabov (approved):

PR #324, opened 2022-06-30
modernize `_compointer_base.value` property

junkmd:
Formerly, the definition of `property` was written in a legacy manner.


junkmd:
I will re-submit all together refactors like this after next version released.

PR #326, opened 2022-06-30
modernize `_Struct_Union_Base` definition

junkmd:
Formerly, `_Struct_Union_Base` has an incomplete constructor.

Given the role of the base class `_Struct_Union_Base`, it should be avoided to assign other than a `StructureBody` instance to the `struct_body` attribute and/or other than a `StructureHead` instance to the `struct_head` attribute.

If there is a situation where a derived class wants to assign instances of a different type to these attributes, it should not be the responsibility of the `_Struct_Union_Base`, and a suitable base class should be defined separately.

### DIFF
--- a/comtypes/tools/typedesc_base.py
+++ b/comtypes/tools/typedesc_base.py
@@ -131,6 +131,10 @@ class StructureBody(object):
 
 class _Struct_Union_Base(object):
     location = None
+    def __init__(self):
+        self.struct_body = StructureBody(self)
+        self.struct_head = StructureHead(self)
+
     def get_body(self):
         return self.struct_body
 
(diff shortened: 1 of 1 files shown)


junkmd:
@vasily-v-ryabov 

This is the last PR I plan to do before the next version is released.
I will continue to refactor and add features after the next version is released.

vasily-v-ryabov (approved):
Nice! Thanks a lot for your big efforts!
I've created milestones 1.1.13 (coming) and 1.1.14 (next). So feel free to add milestone 1.1.14 to next release tasks.

junkmd:
@vasily-v-ryabov 

Thank you for your collaboration!

I was going to start working on 1.1.14 milestone as soon as 1.1.13 was released, but due to my health condition, I will be a little late to start.

vasily-v-ryabov:
1.1.13 is out! When you have time, please review and close some issues. I've closed something that is obviously fixed. Maybe something is missed. And big thanks again!

Issue #330, opened 2022-07-26
`TypeError`  in `client.GetModule("msi.dll")`


junkmd:
I tried `client.GetModule("msi.dll")`, but it failed.

```py
>>> from comtypes.client import GetModule
>>> GetModule("msi.dll")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "...\comtypes\client\_generate.py", line 142, in GetModule
    mod = _create_wrapper_module(tlib, pathname)
  File "...\comtypes\client\_generate.py", line 250, in _create_wrapper_module
    mod = _my_import(fullname)
  File "...\comtypes\client\_generate.py", line 36, in _my_import
    return importlib.import_module(fullname)
  File "...\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 790, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "...\comtypes\gen\_000C1092_0000_0000_C000_000000000046_1033_1_0.py", line 1191, in <module>
    (['out', 'retval'], POINTER(POINTER(Patch)), 'retval')
TypeError: _type_ must have storage info
```

junkmd:
## Assumption of cause

There is `Patch` class definition in `msi.dll`.

`codegenerator` replaces the `Patch` namespace in the generated module with `comtypes.automation.Patch` (defined in `comtypes.patcher.Patch`).

Therefore, type definition and import fails because `comtypes.patcher.Patch` has no storage info.

It will be resolve to replace from `from comtypes.patcher import Patch` and `@Patch` to `import comtypes.patcher` and `@comtypes.patcher.Patch` in `comtypes.automation`.
https://github.com/enthought/comtypes/blob/944e279890e3a4918e7661a8e268e6e2bddf57ae/comtypes/automation.py#L12
https://github.com/enthought/comtypes/blob/944e279890e3a4918e7661a8e268e6e2bddf57ae/comtypes/automation.py#L558-L559

```py
import comtypes.patcher  # or from comtypes import patcher
...
[...]

PR #331, opened 2022-07-26
Fix `TypeError`  in `client.GetModule("msi.dll")`

junkmd:
see #330 and [comment](https://github.com/enthought/comtypes/issues/330#issuecomment-1195419160).

And I added test that ensure #293 was resolved.
(Unused imports in generated modules may be used from other generated modules.)


junkmd:
@vasily-v-ryabov 

I was planning to #326 will be the last PR until the next version is released.
But I found an error and PR'd this for repairing #330 issue.

vasily-v-ryabov (approved):

Issue #333, opened 2022-07-29
Delayed loading of npsupport


bennyrowland:
I have been implementing some COM servers using comtypes, which are generally working very well. The design of the system is such that individual processing operations are provided by each COM server and are launched from a client GUI. In order to keep the GUI thread responsive, the servers each have a second "worker" thread to do their processing, the main thread only handles the COM interactions.

One of the issues that I have been confronting is start-up times for the server application, as this necessarily blocks the client software. This seems to be mainly caused by import times for the various libraries required, and so I have arranged things so that most libraries are imported by the worker thread rather than the main thread - however the most significant remaining library is numpy. I use this for processing, but it is not necessary to have the comtypes numpy interop behaviour. However, because numpy is installed, comtypes insists on importing it at startup, doubling the overall time to launch for a COM server. Do we think it would be possible to modify the way that npsupport is imported, either allowing it to be turned off completely, or even better to allow it to be loaded independently, i.e. import comtypes without it, starting up a server very fast, then import comtypes.npsupport (or call a function like comtypes.enable_np) to switch over to supporting it afterwards?

junkmd:
@bennyrowland 

I also thought a situation like your issue was problematic.

I didn't really have a problem with it because I don't use the `server` feature in my project.

Recently I have been working with restore test code(see #267, #298 and other issues/PRs).
In the process, I noticed the following problem.
- `numpy` is referenced even though it is not a dependency.
  - `comtypes` will be developed as "a pure Python library (without any dependencies)" in the future as well(https://github.com/enthought/comtypes/issues/216#issuecomment-664517594 by @vasily-v-ryabov).
- Behavior of `safearray` and others changes depending on whether `numpy` exists in the environment or not.

As shown in [`numpy-interop`](https://github.com/enthought/comtypes/blob/77cf6b8e2dcedec9a8) in the documentation, using [`safearray_as_ndarray` context manager](https://github.com/enthought/comtypes/blob/77cf6b8e2dcedec9a8bfed0cc03a40aebe5a69b1/docs/source/index.rst#numpy-arrays-as-output-arguments) to cast the return value to `ndarray`.
Similarly, I thought that force to use `safearray_as_ndarray` when using `ndarray` as an argument, but this is not a good idea for backward compatibility since the behavior would be different from the previous one.

Currently, regardless of whether `ndarray` is used or not, `npsupport` is called to check whether `numpy` exists in the environment first.

[...]

PR #337, opened 2022-08-08
Require explicit enabling of numpy interop

bennyrowland:
This PR addresses issue #333, "Delayed loading of npsupport". It does not attempt to load numpy as a library at import time, but only when explicitly requested via the new function `comtypes.npsupport.enable_numpy_interop()` (or by entering the existing context manager `safearray_as_ndarray`). This allows users to control whether (and when) numpy is imported, even if it is installed in their environment.

The current version passes all the previous tests, and has gained one checking that trying to access numpy via npsupport fails until `enable_numpy_interop()` has been called, but probably some further tests are appropriate to test that the right errors are raised at the right times when the function has not been called, so that any users who will be affected by this change get the most helpful error messages. However, best practice testing of this feature may require upgrading of the test environment to allow testing different python envs with and without numpy. There is also the possibility that all numpy interop could be replaced with the PEP3118 Buffer Protocol, removing all numpy related code from comtypes.

### DIFF
--- a/comtypes/npsupport.py
+++ b/comtypes/npsupport.py
@@ -109,10 +109,13 @@ def enable_numpy_interop():
     necessary functions for comtypes to work with ndarrays.
 
     """
+    global HAVE_NUMPY
+    if HAVE_NUMPY:
+        return
+
     global numpy
     import numpy
 
-    global HAVE_NUMPY
     HAVE_NUMPY = True
     global typecodes
     typecodes = _check_ctypeslib_typecodes()
--- a/comtypes/npsupport.py
+++ b/comtypes/npsupport.py
@@ -1,16 +1,13 @@
 """ Consolidation of numpy support utilities. """
 import sys
 
-try:
-    import numpy
-except ImportError:
-    numpy = None
-
-
-HAVE_NUMPY = numpy is not None
-
 is_64bits = sys.maxsize > 2**32
-
+HAVE_NUMPY = False
+com_null_date64 = None
+numpy = None
+datetime64 = None
+VARIANT_dtype = None
+typecodes = {}
 
 def _make_variant_dtype():
     """ Create a dtype for VARIANT. This requires support for Unions, which is
@@ -21,6 +18,7 @@ def _make_variant_dtype():
     Returns None if the dtype cannot be created.
 
     """
+    numpy = get_numpy()
 
     # pointer typecode
     ptr_typecode = '<u8' if is_64bits else '<u4'
@@ -64,6 +62,7 @@ def isndarray(value):
     """
     if not HAVE_NUMPY:
         return False
+    numpy = get_numpy()
     return isinstance(value, numpy.ndarray)
 
 
@@ -98,20 +97,27 @@ def _check_ctypeslib_typecodes():
     return ctypeslib._typecodes
 
 
-com_null_date64 = None
-datetime64 = None
-VARIANT_dtype = None
-typecodes = {}
+def enable_numpy_interop():
+    """ Import the numpy library (if not already imported) and set up the
+    necessary functions for comtypes to work with ndarrays.
 
-if HAVE_NUMPY:
+    """
+    global numpy
+    import numpy
+
+    global HAVE_NUMPY
+    HAVE_NUMPY = True
+    global typecodes
     typecodes = _check_ctypeslib_typecodes()
     # dtype for VARIANT. This allows for packing of variants into an array, and
     # subsequent conversion to a multi-dimensional safearray.
+    global VARIANT_dtype
     try:
         VARIANT_dtype = _make_variant_dtype()
     except ValueError:
         pass
-
+    global datetime64
+    global com_null_date64
     # This simplifies dependent modules
     try:
         from numpy import datetime64
@@ -123,3 +129,18 @@ if HAVE_NUMPY:
             com_null_date64 = datetime64("1899-12-30T00:00:00", "ns")
         except TypeError:
             pass
+
+
+def get_numpy():
+    """ Returns the numpy package if numpy interop is enabled, otherwise raises
+    an import error to remind the user they need to manually enable numpy
+    interop
+
+    """
+    if HAVE_NUMPY:
+        return numpy
+    raise ImportError(
+        "In comtypes>=1.2.0 numpy interop must be explicitly enabled with "
+        "comtypes.npsupport.enable_numpy_interop before attempting to use "
+        "numpy features."
+    )
--- a/comtypes/npsupport.py
+++ b/comtypes/npsupport.py
@@ -63,10 +63,12 @@ def isndarray(value):
     if not HAVE_NUMPY:
         if hasattr(value, "__array_interface__"):
             raise ValueError(
-                f"Argument {value} appears to be a numpy.ndarray, but "
-                "comtypes numpy support has not been enabled. Please "
-                "try calling comtypes.npsupport.enable_numpy_interop() "
-                "before passing ndarrays as parameters."
+                (
+                    "Argument {0} appears to be a numpy.ndarray, but "
+                    "comtypes numpy support has not been enabled. Please "
+                    "try calling comtypes.npsupport.enable_numpy_interop() "
+                    "before passing ndarrays as parameters."
+                ).format(value)
             )
         return False
     numpy = get_numpy()
--- a/comtypes/npsupport.py
+++ b/comtypes/npsupport.py
@@ -2,157 +2,152 @@
 import sys
 
 is_64bits = sys.maxsize > 2**32
-HAVE_NUMPY = False
-com_null_date64 = None
-numpy = None
-datetime64 = None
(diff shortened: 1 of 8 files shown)


junkmd (commented):
AppVeyor CI results show some `SyntaxError` when run on several Python versions.

comtypes/test/test_npsupport.py
```diff
@@ -0,0 +1,314 @@
...
+    """Return the COM refcount of an interface pointer"""
+    import gc
+    gc.collect()
+    gc.collect()
+    o.AddRef()
+    return o.Release()
+
+
+class NumpySupportTestCase(unittest.TestCase):
+    def setUp(self) -> None:
```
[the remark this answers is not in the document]
bennyrowland: Thanks, @junkmd, this was a helpful autocomplete from pycharm to insert the type annotation

comtypes/npsupport.py
```diff
@@ -1,125 +1,156 @@
...
+def isndarray(value):
+    """ Check if a value is an ndarray.
+
+    This cannot succeed if numpy is not available.
+
+    """
+    if not HAVE_NUMPY:
+        if hasattr(value, "__array_interface__"):
+            raise ValueError(
+                f"Argument {value} appears to be a numpy.ndarray, but "
```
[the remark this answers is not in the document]
bennyrowland: Thanks, fixed now.

bennyrowland:
Fixed those issues and tests are now passing on Appveyor.

comtypes/npsupport.py
```diff
@@ -1,125 +1,158 @@
...
+def enable_numpy_interop():
+    """ Import the numpy library (if not already imported) and set up the
+    necessary functions for comtypes to work with ndarrays.
+
+    """
+    global HAVE_NUMPY
+    if HAVE_NUMPY:
+        return
+
+    global numpy
```
vasily-v-ryabov, line 118: Too many `global` keywords usually tells about weak design. Typical way to avoid it is a class with static members. The reason for `global` keyword here is seamless integration of `npsupport.numpy` module object for external user, but internal code uses `get_numpy()` function instead which is also public (not `_get_numpy()`).

comtypes/npsupport.py
```diff
@@ -1,125 +1,158 @@
...
+    try:
+        from numpy import datetime64
+    except ImportError:
+        pass
+    else:
+        try:
+            # This does not work on numpy 1.6
+            com_null_date64 = datetime64("1899-12-30T00:00:00", "ns")
+        except TypeError:
+            pass
```
vasily-v-ryabov, line 143: Too many silent `pass` statements in exception handlers is also not good.

vasily-v-ryabov (changes requested):
The implementation looks too raw. I would think how to improve it using Python best practices.

comtypes/npsupport.py
vasily-v-ryabov: Well, this is legacy code that exists for a while. This comment can be ignored for now.

comtypes/npsupport.py
```diff
@@ -1,125 +1,158 @@
...
+        VARIANT_dtype = _make_variant_dtype()
+    except ValueError:
+        pass
+    global datetime64
+    global com_null_date64
+    # This simplifies dependent modules
+    try:
+        from numpy import datetime64
+    except ImportError:
+        pass
```
vasily-v-ryabov, line 137: Despite it's a legacy code, I would suggest `datetime64 = None` and then do `if datetime64:` check below.
And the whole function could be method `Interop.enable()` of class `Interop`. It would make code design much better. Yes, it adds one more level of names. And since we break backward compatibility, let's break it the most effectively. :)

vasily-v-ryabov:
One more general thing is that only the first import of numpy is slow. All next imports are fast so sometimes maybe it's easier to do `import numpy` instead of calling any function.

comtypes/npsupport.py
bennyrowland: @vasily-v-ryabov I am working on making your proposed changes but actually perhaps we should consider just dropping these lines altogether, datetime64 has been in numpy since at least 1.7, we are now at 1.23 - what versions of numpy do we wish to continue to support?

comtypes/npsupport.py
bennyrowland: I agree that this global is unnecessary, I used that before realising that I needed `get_numpy()` so that I could raise the error if it wasn't enabled. However, there were still a lot of `global`s for the other objects in the top level namespace. I am not completely sure about whether a class approach is better, certainly we don't really want people instantiating such a class, but I have pushed an alternative design with an Interop class that has a single instance (called `interop`) created in the npsupport namespace, thus allowing e.g. npsupport.interop.enable() etc.
A potential improvement to this might be to rename npsupport.py to _npsupport.py and then have in __init__.py `from _npsupport import interop as npsupport` which would drop the extra level of indirection to get to `enable()`?

bennyrowland:
@vasily-v-ryabov you are absolutely right, in my new version I just `import numpy` whenever I need it, but inside the `Interop.numpy` property - this is important because it lets us raise an error if users try to access it without enabling support first.

comtypes/npsupport.py
vasily-v-ryabov: Agree to drop these old numpy versions support. 👍

comtypes/npsupport.py
vasily-v-ryabov: It's a class with static members and static methods. No need to instantiate it. This is good idea to `import interop as npsupport` in `__init__.py`. Would `from comtypes import npsupport` work in this case? If yes, that's perfect!

bennyrowland: So the one reason to instantiate the class is for the `numpy` property - chaining @property and @classmethod is removed in Python 3.11 (see e.g. https://stackoverflow.com/a/64738850) so to allow this to be a property rather than e.g. a `get_numpy()` class method, we need the class to be instantiated. Is there a downside to instantiation in this case?
I will check the import situation but I believe that is how it would work.

vasily-v-ryabov: If there is a static member `interop.numpy` (not a property, just a member), it shouldn't require `@property` or `@classmethod` decorators. I would think about it so:
```python
class interop(object):
    numpy = None

    @staticmethod
    def enable():
        interop.numpy = __import__("numpy")

interop.enable()
print(interop.numpy)
# <module 'numpy' from 'C:\PythonXX\lib\site-packages\numpy\__init__.pyc'>
````

bennyrowland: I can, of course, do it this way - but the key point that I wanted to achieve with the property was to raise an ImportError with a helpful error message about how to enable numpy interop if it wasn't already enabled. If we decide that that message is not important, then I can do it as you suggest.

vasily-v-ryabov (approved):

comtypes/npsupport.py
vasily-v-ryabov: OK, agree. I forgot this point.

vasily-v-ryabov:
Please let me know if anything else is planned for 1.2.0 release. If that's it, I will hopefully publish 1.2.0 next week.

bennyrowland:
@vasily-v-ryabov - sorry, but I don't consider this PR ready to be merged yet - I am still working on renaming things so that `comtypes.npsupport` is actually the `comtypes._npsupport.interop` instance (there are some challenges with testing as we have to un-enable the interop between tests), drop the legacy numpy support and there is also a 1 line bug for some older versions of numpy that I found. Please can you give me a couple of hours to get those things sorted out and push a new version?

vasily-v-ryabov:
@bennyrowland no problem. I have to switch to my main job and will be able to publish 1.2.0 next week anyway. So you have enough time to create one more PR. Thanks for your efforts!

bennyrowland:
@vasily-v-ryabov, thanks. I have one final question for now - how would you feel about adding the `parameterized` package as a test dependency - basically it allows me to write 1 test and have it be expanded into two tests at test time, one where npsupport should be enabled, and one where it shouldn't. This makes it easy to test that the correct Exceptions get raised when interop is disabled, using the same code that is testing correct behaviour when it is enabled, without duplicating all the tests. If you are reluctant to add a dependency just let me know and I will arrange things without the package.

vasily-v-ryabov:
@bennyrowland we're trying to keep `comtypes` itself as pure Python package. Test dependencies are not so strict, but if it is only 1-2 tests I would think about new dependency twice. If it saves a lot of code, let's use it.

bennyrowland:
@vasily-v-ryabov, I do understand, and I figured out how to achieve the same effect with my own test decorator. If you have a look at the new PR, you can see how the tests are set up, but it does feel a bit hacky. If you don't like the way it works at the moment then I also considered setting up the decorator to run the "enabled" and "disabled" tests sequentially within the same test. That is slightly cleaner because it doesn't rely on messing about with frame locals, but it has the disadvantage that you only get one test rather than two. Also, let's move discussion on to the new PR now.

PR #343, opened 2022-08-19
333 optin numpy v2

bennyrowland:
As promised, here are the improvements to the optin numpy PR that didn't quite make it in before the last PR was merged.
`npsupport` has been moved to `_npsupport` so that `_npsupport.interop` can be imported in `__init__.py` as `comtypes.npsupport`. That also means that e.g. `from comtypes import npsupport` is valid. Interop is then enabled via `comtypes.npsupport.enable()`. 

Testing has been improved - tests that should fail with an error if interop is not enabled now have a decorator which calls them twice, once with and once without enabling interop, checking for the error in the disabled branch. I managed to do this without needing a 3rd party library in the end, although it is a little bit hacky.

Numpy support pre 1.7 is now dropped as well.

### DIFF
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -27,6 +27,7 @@ logger.addHandler(NullHandler())
 from ctypes import *
 from _ctypes import COMError
 from comtypes import patcher
+from comtypes._npsupport import interop as npsupport
 
 def _check_version(actual, tlib_cached_mtime=None):
     from comtypes.tools.codegenerator import version as required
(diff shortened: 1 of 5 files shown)


bennyrowland:
Also, @vasily-v-ryabov, we should include these changes into the documentation at https://pythonhosted.org/comtypes/, but that doesn't seem to match the docs folder in the repo, how do the docs work, so I can update those (once the current approach is approved, of course).

vasily-v-ryabov (approved):
Thanks! Probably the pythonhosted docs can't be updated any more. It's a legacy hosting. We should migrate to ReadTheDocs which is recommended way to host the docs. I think it can be done in separate PR.

junkmd:
It is related to #333 and #337

PR #352, opened 2022-09-20
carve out `CodeGenerator.type_name` method to `TypeNamer` class

junkmd:
As a result of previous refactoring, `CodeGenerator.type_name` no longer references `self` except recursively using `self.type_name`.

Therefore, it is now possible to carve out a process into a separate class without having to consider side-effects or callbacks.

Only one concrete class `TypeNamer` is defined in this change.
For use in creating stubs, these may in the future be a base abstract class and multiple derived classes.

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -327,6 +327,23 @@ class CodeGenerator(object):
                     file=output)
         return output.getvalue()
 
+    def _inspect_PointerType(self, t, count=0):
+        if ASSUME_STRINGS:
+            x = get_real_type(t.typ)
+            if isinstance(x, typedesc.FundamentalType):
+                if x.name == "char":
+                    return typedesc.Typedef("STRING", x), count
+                elif x.name == "wchar_t":
+                    return typedesc.Typedef("WSTRING", x), count
+        if isinstance(t.typ, typedesc.FunctionType):
+            return t.typ, count
+        if isinstance(t.typ, typedesc.FundamentalType):
+            if t.typ.name == "void":
+                return typedesc.Typedef("c_void_p", t.typ), count
+        if isinstance(t.typ, typedesc.PointerType):
+            return self._inspect_PointerType(t.typ, count + 1)
+        return t.typ, count + 1
+
     def type_name(self, t):
         # Return a string, containing an expression which can be used
         # to refer to the type. Assumes the 'from ctypes import *'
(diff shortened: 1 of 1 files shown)


comtypes/tools/codegenerator.py
```diff
@@ -186,6 +186,7 @@ def __init__(self, known_symbols=None):
         self.stream = io.StringIO()
         self.imports = ImportedNamespaces()
         self.declarations = DeclaredNamespaces()
+        self.type_name = TypeNamer().type_name
```
vasily-v-ryabov, line 189: It should be `TypeNamer().type_name()` or the method should be defined as property.

vasily-v-ryabov (changes requested):

comtypes/tools/codegenerator.py
junkmd: I don't feel right about methods and properties returning `Callable` objects.

Instead, how about making the `TypeNamer` itself a `Callable` object?

https://github.com/enthought/comtypes/blob/4f3a009c770c6cb1979b910325a8716153fb5bb5/comtypes/tools/codegenerator.py#L1134-L1136
↓
```py
class TypeNamer(object):
    def __call__(self, t):
        # type: (...) -> str
        ...
```

https://github.com/enthought/comtypes/blob/4f3a009c770c6cb1979b910325a8716153fb5bb5/comtypes/tools/codegenerator.py#L183-L189
↓
```py
class CodeGenerator(object):

    def __init__(self, known_symbols=None):
        self.stream = io.StringIO()
        self.imports = ImportedNamespaces()
        self.declarations = DeclaredNamespaces()
        self.type_name = TypeNamer()
        ...
```

junkmd: > I don't feel right about methods and properties returning `Callable` objects.

In other words, it is difficult for me to give proper names to methods and properties that return functions.

`type_name` seems to be returning a string, not a function.

I think it should be `type_name_func` in property case (or `get_type_name_func` in method case).
But their names are ugly for me.

So I first tried to assign the instance method itself.

junkmd:
@vasily-v-ryabov 

I [made `TypeNamer` callable itself](https://github.com/enthought/comtypes/pull/352#discussion_r975032866).

junkmd:
@vasily-v-ryabov

If this PR is merged, the next step is to create derived classes for generating runtime code and for generating type hint stub.

The first would be to make the `External` method an abstract method.
- For runtime code generation, the symbol of wrapper-module (e.g. `_xxxxxxxxxx_xxxxxx_xxxxxx_xxxxxx_xxxxxxxxxxxxxxxxxx.ExternalSymbol`) will be returned.
- For type hint stub generation, the symbol of friendly-named-module (e.g. `friendly.ExternalSymbol`) will be returned.

vasily-v-ryabov (approved):
Ah, OK. I wasn't familiar enough with the code and I thought it was a typo. :) But it was correct code. Anyway thanks for improving it further. :)

junkmd:
@vasily-v-ryabov 

Welcome back!

Passing the instance-bound method is indeed complicated, so I think your suggestion was an appropriate one.

Issue #327, opened 2022-07-17
Generate type stubs at the same time as generating modules


junkmd:
## Abstract of proposal
I wish `client.GetModule` generates `Friendly.pyi` type stub at same time as generating `_xxxxxxxxxxxx_xxxxxxxx_xxxxxxxx_xxxxxxxxxxxx_x_x_x.py` and `Friendly.py` runtime modules.

## Rationale
`comtypes` dynamically creates type-defined modules. This is a feature that other COM manipulation libraries don't have.

However, the type definitions are imported from the wrapper module `_xxxxxxxxxxxx_xxxxxxxx_xxxxxxxx_xxxxxxxxxxxx_x_x_x.py` to the user-friendly module `Friendly.py` as shown below, so all type information is hidden from the type checker.
```py
from comtypes.gen import _xxxxxxxxxxxx_xxxxxxxx_xxxxxxxx_xxxxxxxxxxxx_x_x_x
globals().update(_xxxxxxxxxxxx_xxxxxxxx_xxxxxxxx_xxxxxxxxxxxx_x_x_x.__dict__)
__name__ = 'comtypes.gen.Friendly'
```

In terms of coding usability, it is no different than dynamically calling methods and attributes from a `Dispatch` object, and it is difficult to figure out what API the module has.

Also, the methods and properties of the derived classes of `IUnknown` and `CoClass` in the generated module are defined by the metaclass, so the type checker cannot obtain information on arguments and return values.

I would like to add the feature to generate type stubs in a format according to [PEP561](https://peps.python.org/pep-0561/) at the same time as the runtime module to `client.GetModule` to figure out easily what API the module has.

## When
- After the next version of 1.1.12 is released.

and
- After the refactoring of modules that defines the base class that is imported into the generated module is finished.

and
- After the providing type stubs or adding type hint comments in a format according to [PEP484](https://peps.python.org/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code) into some `static` modules below.
  - `comtypes/__init__.py`
  - `comtypes/automation.py`
  - `comtypes/client/__init__.py`
  - `comtypes/client/_genetate.py`

junkmd:
## Memo for type hint symbols

### Question: Does it require backward compatibility with generic-alias for  builtin classes, related to PEP585?

### Answer: No, "There are currently no plans to remove the aliases from `typing`."

see below;
https://github.com/python/typing/issues/1230
[...]

PR #364, opened 2022-11-02
add type hints to `comtypes/__init__.py`

junkmd:
related to #327 

This PR introduces type hints into statically defined modules.

- Scattered `import`s are now grouped together at the beginning of the code.
  - As for `add_metaclass`, it was moved where "if it was imported from `six`, it would be about here."
- Annotation-only symbols are imported from `typing` using `TYPE_CHECKING`, so `List`, `Tuple` and others are not in the runtime namespace.
- In the case of `_IUnknown_Base`, whose definition exists in runtime, is prevented from being used in `gen` by #360.
- For field annotations, defining `AnnoField` in `hints.pyi`. This can be used only under `if TYPE_CHECKING`.

In order to add more type annotations in the future, there are more symbols imported from `typing` than I use this time.
- I think it is better than having changes in the `import` part over and over.

### DIFF
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -1,16 +1,100 @@
-import types
-import sys
-import os
-
 # comtypes version numbers follow semver (http://semver.org/) and PEP 440
 __version__ = "1.1.14"
 
+import atexit
+from ctypes import *
+from ctypes import _SimpleCData
+from _ctypes import COMError
+import logging
+import os
+import sys
+import types
+
+################################################################
+
+def add_metaclass(metaclass):
+    """Class decorator from six.py for creating a class with a metaclass.
+
+    Copyright (c) 2010-2020 Benjamin Peterson
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy of
+    this software and associated documentation files (the "Software"), to deal in
+    the Software without restriction, including without limitation the rights to
+    use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+    the Software, and to permit persons to whom the Software is furnished to do so,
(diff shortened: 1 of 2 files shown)


vasily-v-ryabov (approved):

kdschlosser:
I know this is merged but there is a problem with how the type annotations are being done. 

Why are Python 2 style annotations being used? For starters these are not going to stay and are going to get removed from Python. The second thing is they are not compatible with `inspect.signature` and no annotations will be available.

The 3rd thing is the use of pyi files is really for c extensions where type hints cannot be read. If it's a python source file then all the type hints should be located in that source file.

vasily-v-ryabov:
@kdschlosser this was my suggestion to keep Py2.7 compatibility for some more time. The current plan is to drop Py2.7 support in the middle of 2023 or maybe even earlier (after some release).

kdschlosser:
It's a lot of work that is going to end up having to be redone in a year or potentially less for type hinting which hasn't been used for how many years now? I think the library will be able to get along for <= 1 year without type hints so they can be done a single time and done properly for Python 3. It is a lot of time spent to end up having it all undone in the not too distant future,

Just my thoughts on it.

junkmd:
In addition to what I planned in #327, I have added typehints in this format to `typeinfo`, `codegenerator` and `typedesc`, but I do not plan to write any more typehints in this format for modules other than these except for modifications.

As for comment annotations, I believe that it is not difficult to convert them to inline annotations since there are tools such as https://github.com/ilevkivskyi/com2ann.

The `Structure` and `Union` fields need to be considered and are being discussed in #327.

Issue #385, opened 2022-11-27
Separate and rename `tools/codegenerator.py` into some modules files


junkmd:
related to enthought/comtypes#327

For coding readability, a similar process should be done for generating runtime code and generating type stub code.
So, the concrete class for runtime code generation and the concrete class for type stub code generation should be implemented from abstracted `tools.codegenerator.CodeGenerator` class.

However, defining these abstract and concrete classes in the same `tools.codegenerator` will result in complicated class names and bloated code in a single module.

Therefore, make `tools.codegenerator` a subpackage with some modules.
- `__init__.py`
- the module for base classes
- the module for runtime code generation
- the module for type stub code generation

The current `tools/codegenerator.py` would be renamed to `tools/codegenerator/runtime.py`.
However, the first PR should only do the following to avoid large diffs (and loss of past commit history when renaming files).

```cmd
...\comtypes>mkdir "comtypes/tools/codegenerator"
...\comtypes>git mv comtypes/tools/codegenerator.py comtypes/tools/codegenerator/runtime.py
...\comtypes>echo from comtypes.tools.codegenerator.runtime import version, name_friendly_module, name_wrapper_module, CodeGenerator> "comtypes/tools/codegenerator/__init__.py"
...\comtypes>type nul > "comtypes/tools/codegenerator/_bases.py"
...\comtypes>type nul > "comtypes/tools/codegenerator/typestub.py"
```

And execute `tree /f comtypes/tools` then results should be follows.
```
...\COMTYPES\TOOLS
│  tlbparser.py
│  typedesc.py
│  typedesc_base.py
│  __init__.py
│
└─codegenerator
        runtime.py
        typestub.py
        _bases.py
        __init__.py
```

The contents of `comtypes/tools/codegenerator/__init__.py` is follows.
```py
from comtypes.tools.codegenerator.runtime import version, name_friendly_module, name_wrapper_module, CodeGenerator
```

**Except for `runtime.py`**, TODO comments might be added.

This should not require any other existing codes to be changed.

junkmd:
This is a GOOD FIRST ISSUE because small amount of change and I want newcomers to be involved in the code generation process.

junkmd:
I close this issue once as we may proceed with the other approach suggested in #327 for type hinting.

PR #389, opened 2022-11-29
add type annotations to `client/__init__.py`

junkmd:
related to #327

There are functions in this module those in some cases return `lazybind.Dispatch`.
However,
- `lazybind` module does not yet have type annotations.
- `lazybind.Dispatch` is an object that implements `__getattr__` and `__setattr__`, the type checker will not error even if specifying a non-existing member.

So for now, each function is annotated to return `Any` if the `interface` is not explicitly specified in the argument.

### DIFF
--- a/comtypes/client/__init__.py
+++ b/comtypes/client/__init__.py
@@ -9,43 +9,47 @@
 #
 ################################################################
 
-import sys, os
 import ctypes
+import logging
+import os
+import sys
 
 import comtypes
 from comtypes.hresult import *
-import comtypes.automation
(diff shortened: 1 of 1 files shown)


junkmd:
@vasily-v-ryabov 

I requested your review because I think it would be healthier for the community to have you review the PRs I submit.

---

This PR is the last of adding type annotations to the statically defined modules mentioned in #327.

As next step, or at the same time, change of the `codegenerator` will be started, including #385.

vasily-v-ryabov (approved):

Issue #392, opened 2022-12-04
Dropping Python2 support schedule and what to do


junkmd:
## Overviews

This package still supports Python 2.7.

However, [it is planned to end support for Python2.7](https://github.com/enthought/comtypes/issues/216#issuecomment-1207158728) in #216.
The timing is [mid-2023 or maybe even earlier](https://github.com/enthought/comtypes/pull/364#issuecomment-1336398580).
And by discussions in this issue, it is planned to [support only non-EOL Pythons(3.7+)](https://github.com/enthought/comtypes/issues/392#issuecomment-1336442774) at drop-2.7 timing.

This issue will close when a version is released that no longer supports Python 2.7.

## The community guidelines up to the dropping Python2 support
- **The [`master` branch](https://github.com/enthought/comtypes/tree/master) would be in maintenance-only-mode. Except for fixing regressions, new contributions to codebase and documentation should be merged into the [`drop_py2` branch](https://github.com/enthought/comtypes/tree/drop_py2).**
- **Related tasks should be labeled [`drop_py2`](https://github.com/enthought/comtypes/labels/drop_py2).**
- **It is assumed that changes to the codebase and documentation will be supporting only Python3.7+.**
- **Please install [`black==22.12.0`]((https://pypi.org/project/black/)), and format the codebase that you changed before submitting PR.**
  - **If you PRed from your forked repo, the GitHub Actions(GHA) will run [`black`](https://pypi.org/project/black/) with [`--check --diff --color`](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#exit-code)**.
  - Auto formatting during PR by GHA or other tools is under consideration.
  - The formatter is only used for development and is not a runtime requirement.

## How to add [`drop_py2` branch](https://github.com/enthought/comtypes/tree/drop_py2) to your forked local and remote repository
1. `git remote add upstream git://github.com/enthought/comtypes.git`
2. `git fetch upstream`
3. `git checkout -b drop_py2 upstream/drop_py2`
4. `git push origin drop_py2`

## To track current progress
See [`drop_py2` labeled issues or PRs](https://github.com/enthought/comtypes/labels/drop_py2) and [the milestone](https://github.com/enthought/comtypes/milestone/4) for current status.

### Summary of changes

Release title should be "1.3.0 Dropping Python 2.7 support and fix ancient bugs"

```
 * Improve error message on non Windows environments. By @CristiFati.
 * Adjust styles of codebase. By @junkmd.
[...]

Issue #229, opened 2021-04-24
Create wheel package of comtypes


nanonyme:
Split from https://github.com/enthought/comtypes/issues/216

Currently new pip is forced to whenever installing this package to first locally create a wheel, then install it. We should be creating and publishing universal wheels. https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels

junkmd:
This is related to #228.

I am thinking be started working on this after [Py3.11 first release](https://devguide.python.org/versions/#supported-versions).

## Assumed changes
https://github.com/enthought/comtypes/blob/cc9a0131edc76bd92073f75e9737aad40cd10c58/setup.py#L7-L11
↓
```py
from setuptools import Command, setup
from setuptools.command.install import install
from setuptools.command.build_py import build_py
```

junkmd:
It would be good to work on this issue at the time of dropping Python 2.7 from support.

(related to #392)

junkmd:
NOTE: this is related to #349.

vasily-v-ryabov:
Current releases (including `1.2.1`, of course) have universal wheels already. Is this issue still relevant or should be closed?

nanonyme:
You're right, this is done.

junkmd:
I recognize a package as a universal wheel available package if it has a `.whl` file in the Built Distribution section of `https://pypi.org/project/NAME/#files`.

- For example, [`pyyaml`](https://pypi.org/project/PyYAML/#files), which is a green package at https://pythonwheels.com/, has Built Distributions, while [`pyspark`](https://pypi.org/project/PyYAML/#files), which is a white package [`pyspark`](https://pypi.org/project/pyspark/#files), which is a white package, does not have a Built Distribution.

https://pypi.org/project/comtypes/#files is satisfied the condition.

nanonyme:
@junkmd it is not possible for projects with Python C extensions to have universal wheels. This includes pyyaml. That kind of projects have to build a per-platform wheel for all targets. Comtypes has no C extensions hence why it's possible here. Your definition for universal wheel is wrong.

junkmd:
Thank you for bringing this to my attention.

I misunderstood the term of "universal wheels".
Upon your clarification, I checked and realized that the definition is documented at https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#wheels

[...]

Issue #228, opened 2021-04-24
Adapt to distutils being removed in 3.12


nanonyme:
Creating an issue about this for tracking. 3.10 release notes mark distutils as deprecated and set a date for removal https://docs.python.org/3.10/whatsnew/3.10.html#distutils

nanonyme:
This links to https://github.com/enthought/comtypes/issues/216 in a way that it might or might not be harder to keep supporting Python 2 once distutils is gone.

junkmd:
It would be good to work on this issue at the time of dropping Python 2.7 from support.

(related to #392)

vasily-v-ryabov:
`1.2.1` uses `setuptools` and supports both Python 3.12 and still Python 2.7. I think the issue can be closed.

PR #394, opened 2022-12-05
Prevent .whl installation (OOTB) on non Win

CristiFati:
*COM* is *Win* specific and doesn't work on other *OS*es, therefore it shouldn't be possible to install the *.whl* on those *OS*es.
More: if it is installed on such *OS*es (although [\[PyPI\]: comtypes](https://pypi.org/project/comtypes) mentions the *Operating System* value as *Microsoft :: Windows* -  but that's easy to be ignored), the error is ambiguous as most people are not aware of the *CTypes* extension module (*\_ctypes.pyd* / *\_ctypes\*.so*). Example: [\[SO\]: cannot import name 'COMError' from '\_ctypes'](https://stackoverflow.com/q/74677364/4788546).

This *PR* addresses the above, using *Python*'s packaging mechanism. Drawbacks:
- The wheel name is long (comtypes-1.1.14-py3.py2-none-win\_amd64.win32.win\_arm64.win\_arm32.win\_ia64.whl*), but that shouldn't be a problem when *PIP* installing most people (again) pay no attention to this detail
- The platform list must be maintained (since the entries are now whitelisted, when a new one will be available (new architecture, *128bit*?) it will have to be added manually)

**Tests**:

- Built the *.whl* (it works on both *Win*, *Nix*). Sample command: `"e:\Work\Dev\VEnvs\py_pc064_03.09_test0\Scripts\python.exe" setup.py bdist_wheel --python-tag py3.py2`
- Installed it successfully on a combination of *Python* versions (*3.\**, *2.7*) and architectures (Intel only (*pc064*, *pc032*)) on *Win* (importing the module afterwards, was successful as well)
    - Failed to install on a *Linux*

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,111 @@
+# Contributing to `comtypes`
+
+Welcome! :smile:
+
+`comtypes` is a community project for developing a lightweight COM client and server framework coded by **pure** `Python` on Windows environments.
+
+We appreciate all contributions, from reporting bugs to implementing new features.
+
+## Table of contents
+- [To keep the community healthy and sustainable :busts_in_silhouette:](#to-keep-the-community-healthy-and-sustainable-busts_in_silhouette)
+- [Reporting bugs :bug:](#reporting-bugs-bug)
+- [Breaking troubles down :flashlight:](#breaking-troubles-down-flashlight)
+- [Suggesting enhancements :sparkles:](#suggesting-enhancements-sparkles)
+- [Contributing to the codebase :open_file_folder:](#contributing-to-the-codebase-open_file_folder)
+- [Contributing to documentation :books:](#contributing-to-documentation-books)
+- [Final words :green_heart:](#final-words-green_heart)
+
+## To keep the community healthy and sustainable :busts_in_silhouette:
+
+If you have to understand all that this package has to offer, it would require an enormous amount of knowledge.
+
+- `Python`:snake:
+- `C`-lang:computer:
+- COM interface, implementation, client and server:door:
+- COM type library functionalities:wrench:
+
+However, there is **no means to say** that you must understand all of these things to be a contributor.
+
+The purpose of this document is to provide a pathway for you to contribute to this community even if you only know a small portion of this package.
+
+Please keep the followings in your mind:
+- :point_right:Please follow the following guidelines when posting a [GitHub Pull Request](https://github.com/enthought/comtypes/pulls) or filing a [GitHub Issue](https://github.com/enthought/comtypes/issues) on this project.
+- :bow:The community may not be able to process and reply to your issue or PR right-away. Participants in the community have a lot of work to do besides `comtypes`, but they would try their best.
+- :book:For code of conduct, please read [Contributor Covenant](https://www.contributor-covenant.org/).
+
+## Reporting bugs :bug:
+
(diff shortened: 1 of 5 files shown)


junkmd:
@CristiFati

Thank you for your suggestions.

Aside from code reviews, I am wondering about something.

The community is currently [planning to end support for Python 2.7 and support for Python 3.7+ only](https://github.com/enthought/comtypes/issues/392).

We are also [considering preparing a branch for development based on after drop-Python2.7 support](https://github.com/enthought/comtypes/issues/327#issuecomment-1336340228).

In this situation, I am wondering if adding version bridges and workarounds that take Python2 into account would not require extra effort to remove them when Python3-only support is starting.

To the other maintainers; what are your thoughts?

@vasily-v-ryabov
@cfarrow
@jaraco

jaraco:
Prioritize support for non-EOL Pythons and then only support older versions where demand exists and resources allow. In my experience on very popular projects, there’s almost zero demand for Python 2.7.

CristiFati:
Of course as things move forward, and it's not good to be left behind. So if wanting to benefit from newer language features (*fstrings*, typing, *match*, ...) support for older versions should be gradually dropped in future versions.

Now, I'm not familiar with the code so I don't know what it takes for keeping code *Python 3.5* compatible for example.

If I was to make a decision, I too would aim only for currently supported *Python* versions, but I'd deprecate the older ones first:
1. Next version should still support everything it supports now, also letting people know (a (big) release note, announcements, other communication channels) that older *Python*s won't be supported
2. Next next version should only support newer *Python*s
    - I'd let at least a couple of months between *#1.* and *#2.*
    - From then on, the process can be repeated (more easily)

Regarding specific demands, I don't know how those are handled, but I'd create a branch starting from a (supported) base state and port any required features into that branch then make a release out of it.

Now regarding this *PR*: since *Python 2.7* is working (installing and importing module) before my changes, I didn't want that **this *PR*** to be the one that breaks support, that's why I added some (clumsy) *Python2* stuff (which I can easily remove if requested).

kdschlosser:
Here is a time table for what the OEL for the different Python version are and are going to be as well.

https://endoflife.date/python


It would only make sense to support Python 3.8 and above in any changes that are going to be made. I believe 3.8 is when a lot of type hinting support was added as well.

junkmd:
Thank you for responses.

My other concern is the code changed in this PR could be further changed because [`distutils`](https://docs.python.org/3/library/distutils.html) that are currently used in `setup.py` are deprecated and [will be removed in Python 3.12](https://github.com/enthought/comtypes/issues/228).

So unless @CristiFati wants to reflect this PR from the next version of `comtypes` right now, I think it would be spending less man-hours to PR to [the branch to add features that are supposed to support only Python3](https://github.com/enthought/comtypes/issues/327#issuecomment-1336340228).

And let's discuss in more appropriate issue #392 about which Python 3.x to start supporting in the future.
In relation to this PR, in `setup.py`, it does not seem necessary to provide workarounds or bridges for each version of Python 3.6 or later (I just did a quick check, so if not, please let me know.).

junkmd:
NOTE: this is related to #229

junkmd:
@CristiFati 

Are you interested in modifying this PR for [`drop_py2` branch](https://github.com/enthought/comtypes/tree/drop_py2), related to #392?

CristiFati:
@junkmd:  Sure.

CristiFati:
Closing as branches have been messed up.

junkmd:
Ok, I am waiting for your next PR.

PR #405, opened 2022-12-13
Prevent .whl installation (OOTB) on non Win

CristiFati:
This is the same as #394, but somehow I got the branches messed up, and ended in a situation that some other commits (already present in the target branch) were included in my *PR*.

Cherry-picked the commits from the old branch in this one.

The *.whl* will **no longer**:

 - Install on non *Win*
 - Install on *Python* < *3.7*
 - Build on *Python 2*

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,9 +88,12 @@ classifiers = [
     'Intended Audience :: Developers',
     'License :: OSI Approved :: MIT License',
     'Operating System :: Microsoft :: Windows',
-    'Programming Language :: Python',
-    'Programming Language :: Python :: 2.7',
-    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.7',
+    'Programming Language :: Python :: 3.8',
+    'Programming Language :: Python :: 3.9',
+    'Programming Language :: Python :: 3.10',
+    'Programming Language :: Python :: 3.11',
+    'Programming Language :: Python :: 3.12',
     'Topic :: Software Development :: Libraries :: Python Modules',
     ]
 
@@ -182,6 +185,7 @@ setup_params = dict(
     ],
 
     platforms=["Windows"],
+    python_requires=">=3.7",
 )
 
 if __name__ == '__main__':
--- a/setup.py
+++ b/setup.py
@@ -68,14 +68,13 @@ class test(Command):
 
 try:
     from wheel.bdist_wheel import bdist_wheel
-    class bdist_wheel_win(bdist_wheel, object):  # @TODO - Python 2 compatible (also super() below) still needed?
+    class bdist_wheel_win(bdist_wheel):
         def get_tag(self):
             win_plats = (  # Copied the list from DistUtils
                 "win-amd64",
                 "win32",
                 "win-arm64",
                 "win-arm32",
-                "win-ia64",  # Py2 only
             )
             tag = super(self.__class__, self).get_tag()
             return tuple(tag[:-1]) + (".".join(wp.replace("-", "_") for wp in win_plats),)
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,24 @@ class test(Command):
             self.failure = self.failure or package_failure
 
 
+try:
+    from wheel.bdist_wheel import bdist_wheel
+    class bdist_wheel_win(bdist_wheel, object):  # @TODO - Python 2 compatible (also super() below) still needed?
+        def get_tag(self):
+            win_plats = (  # Copied the list from DistUtils
+                "win-amd64",
+                "win32",
+                "win-arm64",
+                "win-arm32",
+                "win-ia64",  # Py2 only
+            )
+            tag = super(self.__class__, self).get_tag()
+            return tuple(tag[:-1]) + (".".join(wp.replace("-", "_") for wp in win_plats),)
+
+except ImportError:
+    bdist_wheel_win = None
+
+
 classifiers = [
     'Development Status :: 5 - Production/Stable',
     'Intended Audience :: Developers',
@@ -151,6 +169,7 @@ setup_params = dict(
     cmdclass={
         'test': test,
         'build_py': build_py,
+        'bdist_wheel': bdist_wheel_win,
         'install': post_install,
     },
 
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ try:
                 "win-arm64",
                 "win-arm32",
             )
-            tag = super(self.__class__, self).get_tag()
+            tag = super().get_tag()
             return tuple(tag[:-1]) + (".".join(wp.replace("-", "_") for wp in win_plats),)
 
 except ImportError:
--- a/setup.py
+++ b/setup.py
@@ -181,6 +181,8 @@ setup_params = dict(
         "comtypes.tools",
         "comtypes.test",
     ],
+
+    platforms=["Windows"],
 )
 
 if __name__ == '__main__':


junkmd:
CIs are passed.

<details>
<summary>misc</summary>

https://ci.appveyor.com/project/cfarrow/comtypes/builds/45667988
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/45668785
</details>

junkmd:
@CristiFati 

[`super()` in Python3](https://docs.python.org/3/library/functions.html#super) needs no args. This is a Py2 legacy.
other references:
https://stackoverflow.com/a/576183

I searched `bdist_wheel` usecases, using `super` like this is hard to find.
Please let us know what you refer to.

junkmd:
Out of curiosity, do you know where a cheat sheet of `wheel` module or `bdist_wheel` class?

When I look up `bdist_wheel`, all I see are commands, and it is difficult to find class and function descriptions even in the [`wheel` documentation](https://wheel.readthedocs.io/en/stable/).

CristiFati:
I don't have a clear location in mind (I'm not an expert in that area). There's always the source code (*bdist\_wheel.py*), and for clear tasks (and source is too vast while time is too short) I browse the web.

junkmd:
CIs are passed.

<details>
<summary>misc</summary>

https://ci.appveyor.com/project/cfarrow/comtypes/builds/45681633
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/45682002
</details>

junkmd:
@jaraco 
You are member of [Python Packaging Authority](https://github.com/pypa).
You are the expert in these cases and should know much more than I do.

Do you see any concerns with this PR?

setup.py
```diff
@@ -66,14 +66,34 @@ def run(self):
...
     'Operating System :: Microsoft :: Windows',
-    'Programming Language :: Python',
-    'Programming Language :: Python :: 2.7',
-    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.7',
+    'Programming Language :: Python :: 3.8',
+    'Programming Language :: Python :: 3.9',
+    'Programming Language :: Python :: 3.10',
+    'Programming Language :: Python :: 3.11',
+    'Programming Language :: Python :: 3.12',
```
jaraco, line 96: I used to try to keep up with all of the supported Python versions, but found it to be a maintenance nightmare. Moreover, it's redundant with the `python_requires` directive. Just declare `Python :: 3` and ignore the sub versions.

setup.py
```diff
@@ -66,14 +66,34 @@ def run(self):
...
+                "win-arm64",
+                "win-arm32",
+            )
+            tag = super().get_tag()
+            return tuple(tag[:-1]) + (".".join(wp.replace("-", "_") for wp in win_plats),)
+
+except ImportError:
+    bdist_wheel_win = None
+
+
```
jaraco, line 85: This code feels really brittle. I'd like to see this project and others move away from setup.py altogether. It feels wrong or at least a hack to try to present this module as requiring Windows. It does have a runtime dependency on Windows or at least an implementation of Ctypes that supports the com types, but none of that is part of this package. Moreover, there's nothing special about the wheel that it's not installable on Unix but the sdist is. I'm not confident this unusual, mangled bdist name is going to be supported or useful. It may block releasing to PyPI (though I suspect you've tested this design).

I'm inclined to say no, let's not try to implement platform specificity here. Instead, I'd say to bring this problem to pypa/packaging-problems and explain the use case and request/contribute a solution that any package can avail to declare platform specificity.

setup.py
```diff
@@ -162,6 +183,9 @@ def run(self):
         "comtypes.tools",
         "comtypes.test",
     ],
+
+    platforms=["Windows"],
```
jaraco, line 187: This field is poorly documented, non-standardized, and ineffective. Let's avoid it.

jaraco (changes requested):
Although this change aims to make wheels not installable, it's still not going to prevent their sdist counterparts from being downloaded and installed:

```
$ py -3.10 -m pip-run --use-pep517 git+https://github.com/CristiFati/comtypes@cfati_dev01
Python 3.10.8 (main, Oct 13 2022, 09:48:40) [Clang 14.0.0 (clang-1400.0.29.102)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import comtypes
  File "<stdin>", line 1, in <module>
  File "/var/folders/sx/n5gkrgfx6zd91ymxr2sr9wvw00n8zm/T/pip-run-m5g6fgu_/comtypes/__init__.py", line 7, in <module>
    from _ctypes import COMError
ImportError: cannot import name 'COMError' from '_ctypes' (/opt/homebrew/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/lib-dynload/_ctypes.cpython-310-darwin.so)
```

Let's not accept this and the maintenance burden it adds. Instead, if Cristi would like to pursue a more robust fix, I recommend to contribute a feature to the Python packaging ecosystem that allows any project to provide metadata that would satisfy this need.

CristiFati:
Considering latest remarks, should I close this *PR* as well?

junkmd:
After reading @jaraco's review, I now understand why I could not find sufficient description in the `setup.py` or `setutools` documentation.

It means that `setup.py` is no longer recommended in modern Python packages, and it is recommended to write configuration values in `setup.cfg` or `pyproject.toml`.

I think @CristiFati can contribute to this project by moving the package settings to `setup.cfg` or `pyproject.toml`.

As for introducing a configuration value that would limit installation to Windows only, I think that will come after pypa implements that feature.

I am trying to install an auto-formatter in #418, which may cause conflicts in `setup.cfg` or `pyproject.toml` depending on when we PR.

If @CristiFati can contribute on the migration from `setup.py`, we can discuss them.

junkmd:
I’m sorry. I restored branch.

junkmd:
@CristiFati 

[`test_pip_install.py`](https://github.com/enthought/comtypes/blob/7f7c283531ab388f9c68ed7828f5ab99e55fd45a/test_pip_install.py#L24) and [`appveyor.yml`](https://github.com/enthought/comtypes/blob/02d0a014c6f714fb0c2ef2932378c938342563a5/appveyor.yml#L29-L33) call `setup.py`.

It will probably change to a command like running `pip install https://github.com/USERNAME/comtypes/archive/refs/heads/BRANCHNAME`.

CristiFati:
I didn't work with *pyproject.toml* or *setup.cfg*, I don't know the differences between them and why someone would choose one over the other.

I was aiming to solve a problem (although it's not such a big deal)  that I discovered accidentally on *SO*, (also the place / way that I chose to solve it was a bad one), but apparently there's way more going on, that people like me (outside main maintainers group) are unaware of.

Therefore I propose to close this, and when things will settle (and directions will be clear), "outsiders" will have more chances to contribute.

junkmd:
@CristiFati 

OK. I will close this in the coming days. Thank you for your suggestion.

Followings is that I have been able to find out.

Previous versions of `xlwings` (which used to use this package as a dependency) and the current `pywinauto` use a conditional branch using `sys.platform` in `setup.py` to select the package to be dependencies.

https://github.com/pywinauto/pywinauto/blob/f78302c7df7a71980953d3007ed72de6a2a8e604/setup.py#L59-L74
https://github.com/xlwings/xlwings/blob/0.20.8/setup.py#L15-L32

From these examples, I believe that the responsibility for not allowing the package to be installed that cannot be used with the OS type lies not with pypa or the package itself, but with the project that uses it as a dependency.

So, if you propose to set up the `sys.platform` conditional branch in `setup.py` (or equivalent) of the project that uses `comtypes`, I think that would do what you want to do for a start.

PR #422, opened 2022-12-25
Improve COM error message

CristiFati:
An alternative to #405.

*ComTypes* can end up installed (directly or indirectly) on non *Windows* *OS*es.

Message `ImportError: cannot import name 'COMError' from '_ctypes' (/usr/lib/python3.8/lib-dynload/_ctypes.cpython-38-x86_64-linux-gnu.so)` can be a bit hard to understand for some users (especially if there's no *C* background).

This proposes to make it more clear.

### DIFF
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -7,14 +7,13 @@ from ctypes import _SimpleCData
 try:
     from _ctypes import COMError
 except ImportError as e:
-    e.msg = "\n".join((
-        e.msg,
+    msg = "\n".join((
         "",
         "COM technology not available (maybe it's the wrong platform).",
         "Note that COM is only supported on Windows.",
         "For more details, please check: "
         "https://learn.microsoft.com/en-us/windows/win32/com."))
-    raise e
+    raise ImportError(msg) from e
 import logging
 import os
 import sys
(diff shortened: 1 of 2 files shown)


comtypes/__init__.py
```diff
@@ -4,7 +4,17 @@
 import atexit
 from ctypes import *
 from ctypes import _SimpleCData
-from _ctypes import COMError
+try:
+    from _ctypes import COMError
+except ImportError as e:
+    e.msg = "\n".join((
```
junkmd, line 10: Rewriting the exception message attribute directly is not recommended because the original exception information will be lost.

Raising a new exception while preserving the original exception information is easy with Py3 because the [from clause](https://docs.python.org/3/library/exceptions.html#exception-context) can be used.

```py
…
try:
    from _ctypes import COMError
except ImportError as e:
    msg = …
    raise ImportError(msg) from e
```

Rather than making an effort to achieve the same thing in Py2 for the `master` branch, it would be better to work on the `drop_py2` plan(#392) as it is in progress.

Please change the base of this PR to `drop_py2` and rebase the branch as if it was made from `drop_py2` as well, or remake branch and re-submit PR.

junkmd (changes requested):
Thank you for your suggestion.

CristiFati:
Closing, as being replaced by #423.

Issue #418, opened 2022-12-24
Proposals: Introduce a formatter


junkmd:
I would like to introduce a formatter for codebases in statically defined modules.

I would like to be able to apply a formatter to PRed codebases by GitHub Actions.
I have wanted to introduce it, but was concerned that it would break the old Python syntax.
But on the `drop_py2` plan and its branch, that fear is removed.

Keep in mind, I do not intend to apply this to modules under `comtypes.gen` that are dynamically generated.
Because it would require third-party libraries for runtime dependencies, which goes against the "pure python" intent of this library.
- As much as possible, the `codegenerator` process should be able to generate readable code.
However, if the code generation process itself becomes more complicated because of this, I do not mind if the generated code is a little difficult to read.

junkmd:
@cfarrow 
@jaraco 
@vasily-v-ryabov 
And other community participants

I recommend [`black`](https://pypi.org/project/black/).

GHA setting will be follows.
```
# .../.github/workflows/autofmt.yml
on:
  push:

jobs:
    formatter:
        name: formatter
        runs-on: windows-latest
        strategy:
            matrix:
                python-version: [3.7]
        steps:
          - name: Checkout
            uses: actions/checkout@v3
          - name: Set up Python ${{ matrix.python-version }}
            uses: actions/setup-python@v4
            with:
              python-version: ${{ matrix.python-version }}
          - name: Install black
            run: pip install black==22.12.0
          - name: black
            run: python -m black .
          - uses: stefanzweifel/git-auto-commit-action@v4
            with:
              commit_message: Apply Code Formatter Change
              ref: ${{ github.head_ref }}

```

`black` has fewer settings than other formatters. Thus, there should be less community exhaustion by having to tussle with setting values.

Almost the only debatable setting value would be `line_length`.
The default value for black is [88 characters](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length).

I looked at how some well-known projects set theirs up.

- [`typeshed`](https://github.com/python/typeshed/blob/8671fc5c0f1a5f06ebc0e43c10eddd6714e41d7d/pyproject.toml#L1-L8)
- [`pipenv`](https://github.com/pypa/pipenv/blob/cff5620a478249f487064d5fcd8674978ac6d683/pyproject.toml#L5-L30)
- [`django`](https://github.com/django/django/blob/0bd2c0c9015b53c41394a1c0989afbfd94dc2830/pyproject.toml#L5-L7)
[...]

PR #431, opened 2022-12-30
add workaround for failure on GHA auto-commits

junkmd:
I found out that `stefanzweifel/git-auto-commit-action` will fail if a contributor PRs from the forked repository due to permissions, so I decided to skip it in that case.
So this GHA will be triggered by PRs from the source repository only (In other words, it is not triggered by most contributors' PRs).

I found there seem to be a lot of security problems with doing auto-commit in a public repository like this with GHA alone.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
https://blog.gitguardian.com/github-actions-security-cheat-sheet/

I will consider alternative tools like [pre-commit-ci](https://github.com/apps/pre-commit-ci/) in my free time during the new year vacation.

### DIFF
--- a/.github/workflows/autofmt.yml
+++ b/.github/workflows/autofmt.yml
@@ -6,6 +6,9 @@ jobs:
     formatter:
         name: auto-formatter
         runs-on: windows-latest
+        # Because auto-commit to the branch from a forked repository will fail,
+        # so this GHA will be triggered by PRs from the source repository only.
+        if: github.repository == github.event.pull_request.head.repo.full_name
         steps:
           - name: Checkout
             uses: actions/checkout@v3


junkmd:
CIs are passed.

<details>
<summary>memo</summary>

- pr
  - before GHA: https://ci.appveyor.com/project/cfarrow/comtypes/builds/45802685
  - GHA: https://github.com/enthought/comtypes/actions/runs/3806738976/jobs/6475807610
  - after GHA: none of changes
- others
  - https://ci.appveyor.com/project/pywinauto/pywinauto/builds/45802718

this is related to #418
</details>

Issue #438, opened 2022-12-31
adjust import style and references in `client.dynamic`


junkmd:
Some import styles and references to `module.object` are redundant.
In addition, there are some points that are not a problem at runtime but raise errors or warnings in static analysis.

- For example, `comtypes.COMError` on line 30.
When importing `comtypes.automation`, its "parent" `comtypes` are also loaded, so there is no problem at runtime, but since `comtypes` itself is not imported, static analysis raises an error.

https://github.com/enthought/comtypes/blob/39d9e00e738411076712cb814bfcc2e27b379611/comtypes/client/dynamic.py#L1-L8

https://github.com/enthought/comtypes/blob/39d9e00e738411076712cb814bfcc2e27b379611/comtypes/client/dynamic.py#L22-L33

![image](https://user-images.githubusercontent.com/45822440/210133859-d5532a86-43aa-4fd6-a50a-556442615bd6.png)

Modernize the codebase by removing static analysis errors and import redundancies.

junkmd:
I created this issue to serve as a "model case" for future `good first issue` that I will submit.

I will work on this.

PR #439, opened 2022-12-31
adjust import style and references in `client.dynamic`

junkmd:
see #438

The `IUnknown` is not referenced by other modules, like `dynamic.IUnknown`, so I removed it.
There should be almost no use case to specify this symbol in a roundabout way, which can also be referenced by `comtypes.IUnknown`.

With the change to `from comtypes import automation`, the comment out code that referred to `comtypes.automation` has also been changed.

The `comtypes.typeinfo` has been removed as it is not currently needed at runtime, although it remains in the commented out code. Restoring the comment-outs, the IDE will alert and we should know right away what is missing.

### DIFF
--- a/comtypes/client/dynamic.py
+++ b/comtypes/client/dynamic.py
@@ -1,11 +1,8 @@
 import ctypes
-import comtypes.automation
-import comtypes.typeinfo
-import comtypes.client
-import comtypes.client.lazybind
 
-from comtypes import COMError, IUnknown, _is_object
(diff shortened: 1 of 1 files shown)


junkmd:
CIs are passed.
<details>
<summary>memo</summary>

https://ci.appveyor.com/project/cfarrow/comtypes/builds/45807856
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/45807867
</details>

PR #455, opened 2023-01-07
change type hints from comment annotations to inline annotations in `_memberspec`

junkmd:
Those changes are like as #454.

With reference to [the `CONTRIBUTING.md` in the `typeshed`](https://github.com/python/typeshed/blob/70025c372346288675437fc0bd273db84cc0b3d5/CONTRIBUTING.md), the type aliases were changed to names starting with `_`.

> Type variables and aliases you introduce purely for legibility reasons should be prefixed with an underscore to make it obvious to the reader they are not part of the stubbed API.

While looking at the code, I thought to change `_ComMemberSpec` and `_DispMemberSpec` to [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple), and `_MemberSpec` to [`abc.ABC`](https://docs.python.org/3/library/abc.html#abc.ABC) ([`typing.Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) is only available in Py3.8 and later).
However, since this is an implementation change, I will PR it separately from this PR which changes the type annotation.

### DIFF
--- a/comtypes/_memberspec.py
+++ b/comtypes/_memberspec.py
@@ -15,12 +15,12 @@ from comtypes import _CData
 import comtypes
 
 
-PositionalParamFlagType = Tuple[int, Optional[str]]
-OptionalParamFlagType = Tuple[int, Optional[str], Any]
-ParamFlagType = _UnionT[PositionalParamFlagType, OptionalParamFlagType]
-PositionalArgSpecElmType = Tuple[List[str], Type[_CData], str]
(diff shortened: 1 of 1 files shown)


junkmd:
CIs are passed.

<details>
<summary>memo</summary>

https://ci.appveyor.com/project/cfarrow/comtypes/builds/45865214
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/45865231
</details>

PR #457, opened 2023-01-07
Update `codegenerator` type annotations

junkmd:
As same as #455, with reference to [the `CONTRIBUTING.md` in the `typeshed`](https://github.com/python/typeshed/blob/70025c372346288675437fc0bd273db84cc0b3d5/CONTRIBUTING.md), the type aliases were changed to names starting with `_`.

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -187,9 +187,16 @@ def name_friendly_module(tlib):
 
 ################################################################
 
+_DefValType = _UnionT[lcid, Any, None]
+_IdlFlagType = _UnionT[str, dispid, helpstring]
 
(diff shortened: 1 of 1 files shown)


junkmd:
The `lcid` is defined as a class at firtst and then the namespace is overwritten with its instance later.
Inclusion of such symbols in `Union` will result in an error in runtime.

As a workaround, the `lcid` is to be annotated with quoted strings.

Perhaps renaming the `lcid` class should be required, but I'd like to deal with that in a separate issue.

junkmd:
CIs are passed.

<details>
<summary>memo</summary>

https://ci.appveyor.com/project/cfarrow/comtypes/builds/45866245
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/45866254
</details>

Issue #353, opened 2022-09-21
Iterators in generated files implementing `next()` instead of `__next__()`


jonschz:
The following code was generated on comtypes 1.1.14 using `comtypes.client.GetModule("portabledeviceapi.dll")`:

```python
class IEnumPortableDeviceObjectIDs(IUnknown):
    """IEnumPortableDeviceObjectIDs Interface"""
    [...]
    def __iter__(self):
        return self

    def next(self):
        item, fetched = self.Next(1)
        [...]

    def __getitem__(self, index):
    [...]
```
While trying to do some basic type annotation I realised that a `__next__()` method is missing, hence the return value of `__iter__()` is not an iterator. The best fix which won't break existing code is likely to just automatically add
```python
    def __next__(self):
        return self.next()
```
to every class generated in such a way.

junkmd:
Perhaps the related code is around there.

https://github.com/enthought/comtypes/blob/fed3de69ccc61e37187b4a219daa120a97740663/comtypes/tools/codegenerator.py#L809-L859
- https://github.com/enthought/comtypes/blob/fed3de69ccc61e37187b4a219daa120a97740663/comtypes/tools/codegenerator.py#L799-L807

A com object which has `_NewEnum` will be `Iterator` by metaclass and patching.
https://github.com/enthought/comtypes/blob/fed3de69ccc61e37187b4a219daa120a97740663/comtypes/__init__.py#L410-L433

But a com object with no `_NewEnum` , even if they has name starts with `IEnum...`, would not be `Iterator`.

- For the `__dunder__` method, it would be better to define by a unified way of doing either metaclass or generating-code.
However, this would require man-hours to verify the impact and be large amount of the change.
So "define by a unified way" would be outside the scope of this problem solution.

I think this issue can be resolved by only changing `tools.codegenerator` so that the `__next__` method is defined in the class.

junkmd:
@psisquared2 

Please let me know your code that you expected to work correctly (and terminated abnormally).

I may be able to get some hints from that code as to what test cases would guarantee the behavior.

junkmd:
@psisquared2 

I am planning to add a type-stub generating process to the `codegenerator` module.
However, I do not plan to change the class definition in the runtime code, including the `__dunder__` methods.

So you are free from the worry of code conflictions with your `__next__` method definition and my type-stub generating process when you PR to resolve this issue.

junkmd:
@psisquared2

Can you remind this?

[...]

PR #467, opened 2023-01-13
fix `__next__()` in IEnum code generator

jonschz:
This fixes https://github.com/enthought/comtypes/issues/353, see the discussion there.

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -1029,7 +1029,7 @@ class CodeGenerator(object):
             print("        return self", file=self.stream)
             print(file=self.stream)
 
-            print("    def next(self):", file=self.stream)
+            print("    def __next__(self):", file=self.stream)
             print("        item, fetched = self.Next(1)", file=self.stream)
             print("        if fetched:", file=self.stream)
             print("            return item", file=self.stream)


jonschz:
@junkmd Do you want me to add the test case here as well?

junkmd:
@jonschz 

As a matter of fact, I want a test to ensure that a derived class of `IUnknown` which has name starts with `IEnum` and has methods named "Next", "Skip", "Reset", and "Clone" behaves as an `Iterator`.
This behavior has never been tested before.

However, for the following reasons, changing only this one line should be sufficient for the scope of this PR.

1. Before the change, it was clearly a dead code and the cause of the error in Py3.
2. This does not reduce coverage rate. Appveyor CI does not measure coverage, but I do it in my development environment. This change is a one-line-to-one-line and not adding conditional branches.
3. The discussion in #353 allows us to verify that it behaves as an `Iterator`.
The test code illustrated in #353 will only work if a portable device is connected to the environment.
There are other test cases in this package that are environment-dependent and in some cases do not run, but I want to make as few as possible in the test cases we will create.
It seems to me that there may be other COM objects that are less environment-dependent that would be more suitable than `IEnumPortableDeviceObjectIDs` for `IEnum…` is to be used for the purpose of ensuring its behavior as an `Iterator`.
But I don't know of any such COM object or COM library, so I posted #468 as `help wanted`.

junkmd:
CIs are passed.

<details>
<summary>memo</summary>

https://github.com/enthought/comtypes/actions/runs/3914292684/jobs/6694488613
https://ci.appveyor.com/project/cfarrow/comtypes/builds/45934073
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/45936118
</details>

junkmd (approved):
Thank you!

Issue #345, opened 2022-08-23
Define enumerations in generated modules.


junkmd:
## Motivation

Currently, when we generate a Python module from a COM type library with `client.GetModule`, the stuffs defined as `Enumeration` in the COM type library are defined in Python as aliases for `ctypes.c_int`.

Importing and using them in a production code is rarely done.

They are passed to `STDMETHOD`, `COMMETHOD`, `DISPMETHOD` or `DISPPROPERTY` to define method and property arguments and return values.
It is difficult to add features such as enumerated type member information to these symbols by using classes inherited from `c_int` or by using weird tricks regarding `__dict__`, given the magnitude of impact.

This problem has been left as a comment that marked `XXX` on the `codegenerator`, since this package was born.
https://github.com/enthought/comtypes/blob/cc9a0131edc76bd92073f75e9737aad40cd10c58/comtypes/tools/codegenerator.py#L406-L421

## Summary Examples
This feature is implemented by using the same name but making it an alias of `c_int` in the wrapper module and a subclass of `enum.IntFlag` in the friendly module.

The `.../comtypes/gen/Scripting.py` prior to this change looks something like this.
```py
import comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0 as __wrapper_module__
from comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0 import (
    FileAttribute, DriveTypeConst, TristateMixed, IFileSystem,
    SystemFolder, IFileCollection, StandardStreamTypes,
    _check_version, GUID, IFolderCollection, Dictionary, Directory,
    Remote, Library, StdOut, IOMode,
    __MIDL___MIDL_itf_scrrun_0001_0001_0003, WindowsFolder, VARIANT,
    Alias, helpstring, TristateUseDefault, IScriptEncoder, Archive,
    TextStream, IDrive, IFile, HRESULT, Hidden, StdErr, Folders,
    dispid, typelib_path, IFileSystem3, ReadOnly, ForWriting, StdIn,
    FileSystemObject, ForAppending, Encoder, CDRom, IDriveCollection,
    COMMETHOD, Volume, CompareMethod, BinaryCompare, IDictionary,
    BSTR, Normal, TristateTrue, TristateFalse, Fixed, UnknownType,
    Files, File, _lcid, Tristate, SpecialFolderConst, ForReading,
    ITextStream, __MIDL___MIDL_itf_scrrun_0001_0001_0001, Removable,
[...]

PR #471, opened 2023-01-16
Add unit test for generated IEnum (#468)

### LINKED ISSUE #468: make tests for ensuring `IEnum...` in generated modules behave as a `Iterator`
## **THIS IS FOR [`drop_py2`](https://github.com/enthought/comtypes/tree/drop_py2) PLAN! Please see #392.**

Derived class of `IUnknown` which has name starts with `IEnum` and has methods named "Next", "Skip", "Reset", and "Clone" in generated modules are expected that behave as `Iterator`.

This behavior has never been tested before.

And they not work as expected. This issue was discovered in #353 and will be fixed.

We want to ensure that this behavior is ensured by tests using COM libraries and COM objects that are not environment-dependent.
- The test illustrated in #353 is environment-dependent.

If you have used `IEnum...` in Python2 before, you should have an idea for doing this easier.

Any opinions also would be appreciated.

jonschz:
This fixes #468 by adding a unit test for a generated IEnum... interfaces. Unlike the test discussed in #353, this one does not require any portable device to be connected. As far as I am aware, the DirectShow interfaces should exist on every Windows since 2000.

I also verified that this test fails when you reintroduce the error of #353.

### DIFF
--- a/comtypes/test/test_enum.py
+++ b/comtypes/test/test_enum.py
@@ -1,3 +1,4 @@
+import contextlib
 import unittest as ut
 
 import comtypes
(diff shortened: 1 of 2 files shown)


comtypes/test/test_enum.py
```diff
@@ -0,0 +1,26 @@
+import contextlib
+import unittest as ut
+
+from comtypes import POINTER
+import comtypes.client
+
+
+class Test_Enum(ut.TestCase):
```
junkmd, line 8: As I suggested in #345, I plan to introduce the [`enum`](https://docs.python.org/3/library/enum.html) module.

There are also `CoClass`es those name starts with `Enum`.

To emphasize the difference from these, I think the names should be `class Test_IEnum`, `def test_ienum`, and `test_ienum.py`.

junkmd (changes requested):
LGTM😆! The testing perspectives and ways are very excellent!

junkmd:
CIs are passed.

<details>
<summary>memo</summary>

https://github.com/enthought/comtypes/actions/runs/3934018096/jobs/6728330619
https://ci.appveyor.com/project/cfarrow/comtypes/builds/45955226
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/45955490
</details>

junkmd (approved):
Thank you!

Issue #346, opened 2022-08-28
Is `_cominterface_meta._fix_inout_args` no longer needed now?


junkmd:
`_cominterface_meta._fix_inout_args` is existing since d85a911, 14 years ago.

- https://github.com/enthought/comtypes/blob/cc9a0131edc76bd92073f75e9737aad40cd10c58/comtypes/__init__.py#L716-L722
- https://github.com/enthought/comtypes/blob/cc9a0131edc76bd92073f75e9737aad40cd10c58/comtypes/__init__.py#L588-L666

And the last time it was changed was in `0.6.3` that released by @cfarrow, almost 10 years ago.
https://github.com/enthought/comtypes/releases/tag/comtypes-0.6.3

There are "TODO" comments.
https://github.com/enthought/comtypes/blob/cc9a0131edc76bd92073f75e9737aad40cd10c58/comtypes/__init__.py#L596-L597

I tried to remove the `_fix_inout_args` calling part, and ran `python -m unittest discover -vv -s . /comtypes/test -t comtypes\test` in my Py2.7 and Py3.3-3.10 env, but no error occurred.

- I am not sure whether the current testcases are testing a method with "[in, out] parameters" or not.
- The only commits in the `cpython` repository with messages mentioning `build_callargs` were.
  - https://github.com/python/cpython/commit/f6e7e36c3d3ffa0c537a30253488fca4464ed409

If removing it would not cause any problems, I would like to do that to simplify the code.

junkmd:
As mentioned in #473, we found that although there is some bugs in the current `_fix_inout_args`, it is still a necessary workaround for the `['in', 'out']` parameters.

It is planned to add tests for the `['in', 'out']` parameters.

My question has been resolved and I will close this issue.

PR #473, opened 2023-01-17
Fix old index bug in `call_with_inout`

jonschz:
Another day, another pull request...

I found this rather ancient bug in `call_with_inout`. It triggers whenever a method takes an `['out']` parameter followed by an `['in', 'out']` parameter, which seems to be rather uncommon. My specific problem was with [IPortableDeviceContent::CreateObjectWithPropertiesAndData](https://learn.microsoft.com/en-us/windows/win32/api/portabledeviceapi/nf-portabledeviceapi-iportabledevicecontent-createobjectwithpropertiesanddata).

Before this bugfix it was impossible to call this method without getting an error. Providing four arguments would result in `call takes exactly 4 arguments (5 given)`, and providing the correct three arguments would result in `'LP_c_wchar_p' object cannot be interpreted as an integer`, since the old code would mismatch the expected and actual arguments.

I am fairly certain this change will not break any correct existing code, but there is a chance it may break some code containing a workaround for this bug.

Fun fact: `git blame` shows that this code (including the bug) was pushed 14 years ago in https://github.com/enthought/comtypes/commit/8776b1bb3cda79ae9dccb47348f6138cdfde3b90 and never modified.

I also cleaned up the code a little bit which does not affect the logic at all, but increases readability.

**One open question**
We should think about how to handle the case when a parameter has neither `'out'` nor `'in'` is specified. As far as I am aware, this does not happen in the code generator, but it if did, it would likely cause this code to malfunction. One idea would be to raise an exception, the other idea would be to log a warning and then explicitly set the parameter to be `'in'`. I think the old code does the latter implicitly.

One could also consider merging this bugfix into `master` since it is a bugfix for existing python2 code.


junkmd (commented):
Thank you for your contribution!

It is a happy situation for OSS maintainers such as me to the community has contributors who are willing to investigate defects and PR proper level of granularity.

I think this problem is getting to the crux of this project.

The `_fix_inout_args` may have never been tested by boundary values.

As I mentioned in #346, the non-skipped existing tests succeeded with or without the `_fix_inout_args`. I also did not encounter any problems with the projects I am involved with that use `comtypes`.

I think it is necessary another PR adding the tests with three cases.

- Test the behavior of methods that can work without the `_fix_inout_args`
- Test the behavior of methods that can work with the current `_fix_inout_args`.
- Test the behavior of methods that can work with #473 changes
  - Until the production code is changed, this test must be decorated by the [`unittest.expectedFailure`](https://docs.python.org/3/library/unittest.html#unittest.expectedFailure).

Once the test-only PR is merged, you can simply rebase this branch.

This is my opinion and there may be other better ways to do.

Any opinions also would be appreciated.

vasily-v-ryabov:
Thnaks a lot! I'd suggest to backport it to current master branch as well as a separate PR. From point of being explicit, we should raise an exception in case `in`, `out` or `inout` specification is not set. I'm pretty sure this situation should never happen, but never say "Never!". :) Also something like `assert_never` is additionally useful from static typing point of view (for `drop_py2` branch).

jonschz:
I added a unit test here which fails on the unmodified `_memberspec.py`. I also updated the error handling and made the specific routine type safe, revealing another potential edge case which I handled. In my opinion it would be fine to just merge both the fix and the test together, but I could also make a separate pull request for the test only if you prefer that.

junkmd:
OK, adding tests and changing the production code  in one PR would be fine.

However, it is better to use the actual com library for testing.

Please show us the code that reproduces the problem you encountered while using `IPortableDeviceContent.CreateObjectWithPropertiesAndData`.

I am guessing that `IPortableDeviceContent.CreateObjectWithPropertiesAndData` will be environment dependent in behavior, like other portabledevice api.

However, just as you were able to create the IEnum test from the DirectShow api, members of the community may be able to come up with a more suitable test from your code.

jonschz:
I have temporarily added a test [here](https://github.com/enthought/comtypes/blob/b52c373e274455d978944847450b51168455512b/comtypes/test/test_portabledevice.py) to be removed later. It does not create new files on my device, but I can't guarantee that it never does so.

comtypes/test/test_portabledevice.py
```diff
@@ -0,0 +1,146 @@
+ # so we don't crash when the generated files don't exist
+from __future__ import annotations
```
junkmd, line 2: This `__future__` is not used in other modules in `comtypes`.
So you should use string literals as annotations instead of this.

comtypes/test/test_portabledevice.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def newGuid(*args: int) -> comtypes.GUID:
+    guid = comtypes.GUID()
+    guid.Data1 = c_uint32(args[0])
+    guid.Data2 = c_uint16(args[1])
+    guid.Data3 = c_uint16(args[2])
+    for i in range(8):
+        guid.Data4[i] = c_int8(args[3 + i])
+    return guid
```
junkmd, line 26: Instead of using the `newGuid` function, generate the GUID by passing a string to the constructor that the GUID originally has.

comtypes/test/test_portabledevice.py
```diff
@@ -0,0 +1,146 @@
...
+        WPD_OBJECT_CONTENT_TYPE = PropertyKey(
+            0xEF6B490D, 0x5CD8, 0x437A, 0xAF, 0xFC, 0xDA, 0x8B, 0x60, 0xEE, 0x4A, 0x3C, 7)
+        WPD_OBJECT_SIZE = PropertyKey(
+            0xEF6B490D, 0x5CD8, 0x437A, 0xAF, 0xFC, 0xDA, 0x8B, 0x60, 0xEE, 0x4A, 0x3C, 11)
+        WPD_OBJECT_ORIGINAL_FILE_NAME = PropertyKey(
+            0xEF6B490D, 0x5CD8, 0x437A, 0xAF, 0xFC, 0xDA, 0x8B, 0x60, 0xEE, 0x4A, 0x3C, 12)
+        WPD_OBJECT_PARENT_ID = PropertyKey(
+            0xEF6B490D, 0x5CD8, 0x437A, 0xAF, 0xFC, 0xDA, 0x8B, 0x60, 0xEE, 0x4A, 0x3C, 3)
+        folderType = newGuid(0x27E2E392, 0xA111, 0x48E0,
+                             0xAB, 0x0C, 0xE1, 0x77, 0x05, 0xA0, 0x5F, 0x85)
```
junkmd, line 77:
```suggestion
        folder_type = comtypes.GUID("{27E2E392-A111-48E0-AB0C-E17705A05F85}")
```

It should be applied to other `newGuid`s.

comtypes/test/test_portabledevice.py
```diff
@@ -0,0 +1,146 @@
...
+    return guid
+
+
+def PropertyKey(*args: int) -> _Pointer[port_api._tagpropertykey]:
+    assert len(args) == 12
+    assert all(isinstance(x, int) for x in args)
+    propkey = port_api._tagpropertykey()
+    propkey.fmtid = newGuid(*args[0:11])
+    propkey.pid = c_ulong(args[11])
+    return pointer(propkey)
```
junkmd, line 35:
```suggestion
def PropertyKey(fmtid: comtypes.GUID, pid: int) -> "_Pointer[port_api._tagpropertykey]":
    propkey = port_api._tagpropertykey()
    propkey.fmtid = fmtid
    propkey.pid = c_ulong(pid)
    return pointer(propkey)
```

It is better to give an explicit name and type to the argument than to add assert to inspect the argument.

comtypes/test/test_portabledevice.py
```diff
@@ -0,0 +1,146 @@
...
+        self.device = comtypes.client.CreateObject(
+            port_api.PortableDevice().IPersist_GetClassID(),
+            clsctx=comtypes.CLSCTX_INPROC_SERVER,
+            interface=port_api.IPortableDevice,
+        )
+        self.device.Open(list(dev_ids)[0], info)
+
+    def test_EnumObjects(self):
+        WPD_OBJECT_NAME = PropertyKey(
+            0xEF6B490D, 0x5CD8, 0x437A, 0xAF, 0xFC, 0xDA, 0x8B, 0x60, 0xEE, 0x4A, 0x3C, 4)
```
junkmd, line 67:
```suggestion
        WPD_OBJECT_NAME = PropertyKey(comtypes.GUID("{EF6B490D-5CD8-437A-AFFC-DA8B60EE4A3C}"), 4)
```
It should be applied to other `PropertyKey`s.

comtypes/test/test_portabledevice.py
```diff
@@ -0,0 +1,146 @@
...
+        functionalType = newGuid(
+            0x99ED0160, 0x17FF, 0x4C44, 0x9D, 0x98, 0x1D, 0x7A, 0x6F, 0x94, 0x19, 0x21)
+
+        content = self.device.Content()
+        properties = content.Properties()
+        propertiesToRead = comtypes.client.CreateObject(
+            port_types.PortableDeviceKeyCollection,
+            clsctx=comtypes.CLSCTX_INPROC_SERVER,
+            interface=port_api.IPortableDeviceKeyCollection,
+        )
```
junkmd, line 87: Perhaps it would be easier to read if it was simply `key_collection`.

junkmd (changes requested):
My guess is that you will continue to improve this test code (even if in your local env) and post issues about technical problems that occurred there.

Therefore, please keep in mind that this code will be shared with us again in the future, so please make sure that it is more readable.

jonschz:
Oh, a misunderstanding - I was under the impression that you merely wanted to see some example test code and didn't want to integrate this test into production code (especially because I can't guarantee that it doesn't modify content on portable devices). I will update the code.

comtypes/test/test_portabledevice.py
jonschz: The reason I did it this way is because it allows easy copy-paste from the Windows header files, eliminating a possible source of errors. I can of course change that.

comtypes/_memberspec.py
```diff
@@ -175,54 +175,65 @@ def call_with_inout(self, *args, **kw):
...
+            is_positional = param_index < len(args)
+            if dir_in and dir_out:
                 # This is an [in, out] parameter.
                 #
                 # Determine name and required type of the parameter.
                 name = info[1]
                 # [in, out] parameters are passed as pointers,
                 # this is the pointed-to type:
-                atyp = argtypes[i]._type_
+                atyp: type[_CData] = getattr(argtypes[i], "_type_")
```
junkmd, line 193: I think that this line is not covered by the current test.

Because Py <= 3.8, the `type` was not ["Type Hinting Generics In Standard Collections"](https://peps.python.org/pep-0585/), which should result in an error if this line is executed.

comtypes/test/test_portabledevice.py
```diff
@@ -0,0 +1,129 @@
...
+                for x in enumobj:
+                    objId = findDir(x, level - 1)
+                    if objId:
+                        return objId
+                return None  # not in this part of the tree
+            else:
+                return None
+
+        # find a directory 2 levels deep, because you can't write on the top ones
+        parentId = findDir("DEVICE", 2)
```
junkmd, line 93: The variable name should be the PEP-8ish `parent_id`.

The same applies to other variable names.

junkmd (changes requested):
I would agree that add a test using portabledevice to the project, under the following conditions ...

- Assume the target device is a sandbox consisting of a few folders and a few very small files.
- Use `os.path`, `glob`, `pathlib`, etc. to check if the target device has the same directory structure as the expected sandbox or skip the test if not.
- Eliminate the using conditional branches such as inside of `findDir` that are not used to skip or fail tests.
- Make assertions that inspect contents of the target device.
- If doing something like `RemoteWrite`, create a temporary folder using `tempfile` or something similar and write to it.
- If possible, assert after the test that the sandbox is not different from before the test.

Existing this test is no matter, however, be that as it may, I think it is necessary to test `_fix_inout_args` in another environment-independent way, since it is difficult to always require such a sandbox in developer and CI environments.

junkmd:
As for `RemoteWrite`, if it can be verified by tests like #474, I think not need to take the risk of testing it with portabledevice.

comtypes/_memberspec.py
jonschz: I was very surprised to find that this line is indeed covered by the test, but it still does not crash in Python 3.7. In any case, I will fix it.

jonschz:
> However, it is better to use the actual com library for testing.

Would you reconsider your stance on this matter, since it appears to be hard to catch all edge cases of `_fix_inout_args`? I have some unpublished improvements for `test_inout_args.py` that give an almost 100% test coverage of `_fix_inout_args`.

junkmd:
Thanks a lot for your contribution!

Indeed, I too felt that it would be difficult to cover all edges in the actual COM library.
Sorry for pushing you to do this.

Currently you are trying to cover the edges implicitly with `subTest`, but it would be better to name the tests explicitly for each test perspective.
Then, for each test, you can comment "this edge is assumed to be a ham and is also covered by `test_foo.py`", "this edge is assumed to be a lettuce but is not covered by the test with the actual com library", etc.

Then, you can use a mock to assert how the function was actually called.
Below is an example of testing a higher-order function that adds very simple arguments.

```py
import functools as ft
from typing import Any, Callable
import unittest as ut
from unittest.mock import MagicMock


def add_a_eqaul_one_arg(func: Callable[..., Any], *args, **kwargs):
    @ft.wraps(func)
    def wrapper(*args, **kwargs):
        return func(*args, **kwargs, a=1)

    return wrapper

class Test(ut.TestCase):
    def test(self):
        m = MagicMock()
        new_m = add_a_eqaul_one_arg(m)
        new_m(4, 3, b=2)
        m.assert_called_once_with(4, 3, a=1, b=2)

```
I think that such a test is necessary if it cannot be provided actual stuffs.

jonschz:
I have rewritten and expanded the unit test, feel free to take a look when you have time.

Surprisingly, a different test had some global side effects despite it being skipped, breaking my new test. I made some modifications there as well.

Regarding the lines
https://github.com/enthought/comtypes/blob/dac3a12251628049e7f90e73ebb774295dd898a1/comtypes/_memberspec.py#L186-L188
I was unable to find a practical use case where they would be called, and none of the existing tests cover them either. Any advice would be appreciated. If it turns out that they don't appear to be used at all, we could add a comment explaining the situation but leave the code in for backwards compatibility.

junkmd (changes requested):
Your thoughtfulness is excellent.
I am using my free time to this PR because your proposal deals with how methods should be invoked with what args, that is also adjacent to static typing.

I am still concerned about the large number of `if` clauses in the `TestEntry`([This SO explains why avoiding such things](https://stackoverflow.com/questions/27880631/why-should-you-avoid-conditional-logic-in-unit-tests-and-how)).
In particular, `mock_function` reproduces the behavior of `WINFUNCTYPE`(is my recognition right?).
But we need to ensure the correctness of this.

Rather, I prefer to create and inspect a mock that assumes the objects returned by the "raw" `WINFUNCTYPE`.
Following test code is hastily prepared, so it may be flawed in some ways, but please point them out to me if so.

```py
from comtypes import COMMETHOD
from ctypes import HRESULT

WSTRING = c_wchar_p
...

class Test_FixInOutArgs(ut.TestCase):
    def test_PortableDeviceApiLib_CreateObjectWithPropertiesAndData(self):
        spec = COMMETHOD(
            [],
            HRESULT,
            "CreateObjectWithPropertiesAndData",
            (["in"], POINTER(IUnknown), "pValues"),  # IPortableDeviceValues
            (["out"], POINTER(POINTER(IUnknown)), "ppData"),  # IStream
            (["in", "out"], POINTER(c_ulong), "pdwOptimalWriteBufferSize"),
            (["in", "out"], POINTER(WSTRING), "ppszCookie"),
        )
        m = MagicMock()
        fixed = _fix_inout_args(m, spec.argtypes, spec.paramflags)
        p_val = MagicMock(
            spec=POINTER(IUnknown), name="POINTER(IPortableDeviceValues)"
        )()
        pp_data = MagicMock(name="POINTER(POINTER(IStream))")()
        buf_size = MagicMock(spec=POINTER(c_ulong), name="POINTER(c_ulong)")()
        cookie = MagicMock(spec=POINTER(WSTRING), name="POINTER(c_wchar_p)")()
        m.return_value = (pp_data, ..., ...)
        ret_val = fixed(p_val, buf_size, cookie)
        self.assertEqual(ret_val, [pp_data, 1, None])
        m.assert_called_once()
        (m_1st, m_2nd, m_3rd), m_kw = m.call_args
        self.assertEqual(m_1st, p_val)
        self.assertEqual(m_2nd, buf_size)
        self.assertIsInstance(m_3rd, c_ulong)
        self.assertEqual(list(m_kw), ["ppszCookie"])  # ensure length and keys
        self.assertIsInstance(m_kw["ppszCookie"], c_wchar_p)
```
This PR deals with a difficult problem for test-driven development.
My code is only a suggestion, and there should be many different test perspectives.

Any opinions also would be appreciated.
Thank you.

junkmd:
> leave the code in for backwards compatibility.

Please do so for now.

jonschz:
Thank you for your thoughts and feedback. I went on to implement some of them, with the following comments:
- I have simplified and improved the mock functions based on your suggestions, in particular regarding the return types. This made lots of if-statements obsolete.
- It turns out not to be the best idea to pass `MagicMock()`-argument into `fixed()`, because there is conditional logic in `_fix_inout_args()` that does not work well with mock parameters.
- One detail you missed in your draft was to either use `comtypes.instancemethod` or pass in `self` in `fixed(self, ...)`. All the parameters were hence shifted one to the left, and the cookie got auto-generated.
- While writing the tests I encountered what I believe to be inconsistent logic in a different part of `_fix_inout_args()` which does not appear to have any real effect. I have put some temporary comments into `_memberspec.py` describing the issue, and I would value your opinion on this matter. My intuition is not to change the existing code and work around it in the tests, but it will make the tests less tidy.
- Feel free to modify code directly in the branch of this pull request, I have given the necessary permissions.

junkmd:
Thank you for your cooperation.

> * One detail you missed in your draft was to either use `comtypes.instancemethod` or pass in `self` in `fixed(self, ...)`. All the parameters were hence shifted one to the left, and the cookie got auto-generated.

Oh, I missed that...
Thank you.

> * Feel free to modify code directly in the branch of this pull request, I have given the necessary permissions.

I would like to add a test like `test_CreativeObjectWithPropertiesAndData` to your code.
I will post a comment on this PR when the modification is complete.

> * While writing the tests I encountered what I believe to be inconsistent logic in a different part of `_fix_inout_args()` which does not appear to have any real effect. I have put some temporary comments into `_memberspec.py` describing the issue, and I would value your opinion on this matter. My intuition is not to change the existing code and work around it in the tests, but it will make the tests less tidy.

I will leave this part as is for now. I would like to hear other maintainers' opinions on this part.

junkmd:
For now, I have added tests for the cases where...
- `CreateObjectWithPropertiesAndData` is called in various ways
- error occurs

Please check it out.

The following is my opinion.

- Although much less than before, it is still difficult to determine that tests are legitimate if there are `if` clauses in the `TestEntry`.
The arguments being passed, the values being returned, and their combinations should be explicit.
- Instead of using `permutations` to test all patterns, it is better to write code explicitly to test a few of them.
- I am wondering if the `IRecordInfo::GetFieldNames` and `ITypeLib::IsName` memberspecs are appropriate to test.
In `comtypes`, the methods named "GetFieldNames" in `IRecordInfo` and "IsName" in `ITypeLib` wrap low-level methods and and override the namespaces. So the high-level methods generated by the metaclass are not invoked (unless we go around them).
  - https://github.com/enthought/comtypes/blob/532c3996e73866009249e02981d3ead9ae58d474/comtypes/typeinfo.py#L259-L272

  - https://github.com/enthought/comtypes/blob/532c3996e73866009249e02981d3ead9ae58d474/comtypes/typeinfo.py#L876-L889

  - https://github.com/enthought/comtypes/blob/532c3996e73866009249e02981d3ead9ae58d474/comtypes/typeinfo.py#L483-L490

  - https://github.com/enthought/comtypes/blob/532c3996e73866009249e02981d3ead9ae58d474/comtypes/typeinfo.py#L544-L550

jonschz:
Thank you for the contributions and efforts!

I corrected some minor issues I found. Concerning the points you made:

> * Although much less than before, it is still difficult to determine that tests are legitimate if there are `if` clauses in the `TestEntry`.
>       The arguments being passed, the values being returned, and their combinations should be explicit.

I agree that some more cleanup and documentation is needed, but I don't think you can get around all of them if you want the mock function to correctly emulate the behaviour of the to-be-fixed function. Depending on the number of `out` or `inout` parameters, it returns an `HRESULT`, a single result, or a tuple of results. Regarding `run_test`, there is more automatic and less explicit testing, which I see as an addition and not an alternative to the other more explicit tests you added.

> * Instead of using `permutations` to test all patterns, it is better to write code explicitly to test a few of them.

I would argue that we should leave the permutations in, in addition to adding some more real-life scenarios. After all, it was a rather rare permutation that was causing the problem in the first place, and even if there had been tests in the past, they would likely have missed this error unless they would have tested permutations.

> * I am wondering if the `IRecordInfo::GetFieldNames` and `ITypeLib::IsName` memberspecs are appropriate to test.

Feel free to replace them by any other pair of methods, those were the first two I came across and one of them had a name of `None`, which I found to be an interesting edge case.

Still TBD is a test case with only one `inout` and no further `out` arguments, which is related to the `expectedFailure` and the edge case I discussed here:

> * While writing the tests I encountered what I believe to be inconsistent logic in a different part of `_fix_inout_args()` which does not appear to have any real effect.

After that, we can remove the whole `TestEntry` code.

junkmd:
Thank you for your response.

I have an idea for the tests instead of using `permutations` and `if` statements.
I will write and push it in my free time.

Luckily, I was recently working on implementations using `@typing.overload` and `*args, **kwargs` in my project, so I am familiar with tests dealing with breaking down and re-passing such arguments.

At that time, I wrote tests in a verbatim way using `pytest` and `pytest.mark.parametrize`.

I will try to write something similar using `unittest` and `subTest`.

junkmd:
For leaving the arguments permutations while making less using `if` statements, I wrote code for 24 parameter patterns(permutations of 4 elements) in verbatim way with `unittest` and `subTest`.

I am still struggling with `real_world_examples`.
Do you have any suggestions?

junkmd:
I came up with a test for `real_world_examples`.
I will work for it in my free time.

jonschz:
Oh, I am working on it right now. Will push soon, maybe you can still add your tests.

jonschz:
Here are the new real life tests, feel free to add your own ones.

- Coincidentally, I managed to get 100 % test coverage on `_fix_inout_args()`.

- What I also found was that running the tests on the original, bugged code in `_memberspec.py` crashes the python interpreter silently. The culprit is the following function call:
  ```python
        ctypes.c_wchar_p(5)__ctypes_from_outparam__()
  ```
  I was wondering if the CI handles this kind of problem correctly.
- While looking for real-life test subjects I came across a function with a parameter that has neither `'out'` nor `'in'` specified. I will investigate and open a separate issue.

jonschz:
> I'm pretty sure this situation should never happen, but never say "Never!"

@vasily-v-ryabov True words - I have found some examples. See e.g. `msvidctl.dll`, `IMFAttributes.GetString()`:
```python
COMMETHOD(
    [],
    HRESULT,
    'GetString',
    (
        [],
        POINTER(comtypes.gen._00020430_0000_0000_C000_000000000046_0_2_0.GUID),
        'guidKey',
    ),
    (['out'], WSTRING, 'pwszValue'),
    ([], c_uint, 'cchBufSize'),
    (['in', 'out'], POINTER(c_uint), 'pcchLength')
)
```
I dug a little deeper and found that this is not an issue of the code generator. From `Mfobjects.idl`:
```C
HRESULT GetString(
        REFGUID                           guidKey,
        [out, size_is(cchBufSize)] LPWSTR pwszValue,
        UINT32                            cchBufSize,
        [in, out, ptr] UINT32*                     pcchLength
        );
```
There are other examples in this file as well. To me, the most reasonable decision would be to treat missing `in` and `out` as `in`, like `_memberspec.py` did previously, but I would value your opinions.

comtypes/test/test_inout_args.py
```diff
@@ -0,0 +1,485 @@
...
+
+        self.assertEqual(
+            ret_val,
+            [
+                self.pp_data,
+                orig_kw["pdwOptimalWriteBufferSize"].value,
+                orig_kw["ppszCookie"].value,
+            ],
+        )
+
```
junkmd, line 234:
```suggestion


@ut.expectedFailure
class Test_ArgsKwargsCombinations_2(ut.TestCase):
    def setUp(self):
        # a memberspec of `PortableDeviceApiLib.IPortableDeviceContent`
        spec = comtypes.COMMETHOD(
            [],
            HRESULT,
            "CreateObjectWithPropertiesOnly",
            (["in"], POINTER(comtypes.IUnknown), "pValues"),  # IPortableDeviceValues
            (["in", "out"], POINTER(WSTRING), "ppszObjectID"),
        )
        self.orig = MagicMock()
        self.fixed = _fix_inout_args(self.orig, spec.argtypes, spec.paramflags)

    def test_positionals_only(self):
        self_ = MagicMock(name="Self")
        p_val = MagicMock(spec=POINTER(IUnknown))()
        objid = "foo"
        ret_val = self.fixed(self_, p_val, objid)

        self.assertEqual(ret_val, objid)
        self.orig.assert_called_once()
        (orig_0th, orig_1st, orig_2nd), orig_kw = self.orig.call_args
        self.assertIs(orig_0th, self_)
        self.assertIs(orig_1st, p_val)
        self.assertIsInstance(orig_2nd, WSTRING)
        self.assertEqual(orig_2nd.value, objid)
        self.assertEqual(orig_kw, {})

    def test_keywords_only(self):
        self_ = MagicMock(name="Self")
        p_val = MagicMock(spec=POINTER(IUnknown))()
        objid = "bar"
        ret_val = self.fixed(self_, ppszObjectID=objid, pValues=p_val)

        self.assertEqual(ret_val, objid)
        self.orig.assert_called_once()
        (orig_0th,), orig_kw = self.orig.call_args
        self.assertIs(orig_0th, self_)
        self.assertIs(orig_kw["pValues"], p_val)
        self.assertIsInstance(orig_kw["ppszObjectID"], WSTRING)
        self.assertEqual(orig_kw["ppszObjectID"].value, objid)

    def test_mixed_args(self):
        self_ = MagicMock(name="Self")
        p_val = MagicMock(spec=POINTER(IUnknown))()
        objid = "baz"
        ret_val = self.fixed(self_, p_val, ppszObjectID=objid)

        self.assertEqual(ret_val, objid)
        self.orig.assert_called_once()
        (orig_0th, orig_1st), orig_kw = self.orig.call_args
        self.assertIs(orig_0th, self_)
        self.assertIs(orig_1st, p_val)
        self.assertIsInstance(orig_kw["ppszObjectID"], WSTRING)
        self.assertEqual(orig_kw["ppszObjectID"].value, objid)

    def test_omitted_arguments_autogen(self):
        self_ = MagicMock(name="Self")
        p_val = MagicMock(spec=POINTER(IUnknown))()
        self.orig.return_value = "qux"
        ret_val = self.fixed(self_, pValues=p_val)

        self.orig.assert_called_once()
        (orig_0th,), orig_kw = self.orig.call_args
        self.assertEqual(ret_val, orig_kw["ppszObjectID"].value)
        self.assertIs(orig_0th, self_)
        self.assertEqual(set(orig_kw), {"pValues", "ppszObjectID"})
        self.assertIs(orig_kw["pValues"], p_val)
        self.assertIsInstance(orig_kw["ppszObjectID"], WSTRING)
        self.assertEqual(orig_kw["ppszObjectID"].value, WSTRING().value)

```
I came up with a new test and wrote it, but looking back after you added the new test, this is not `real_world_examples`, but rather a test that covers the case where there is `['in', 'out']` but `['out']` is not in the memberspec . And this test will not work without resolving the problem that you explained below, so this testcase is just suggestion.

https://github.com/enthought/comtypes/blob/3b72578ff58af2b1e984304cfac1e7bf41fd1702/comtypes/_memberspec.py#L238-L257

I understood clearly your explanation after writing my own test code.
Indeed, It seems to me this is a wierd behavior.

junkmd:
I thought I made a normal comment, but it turned out to be a review comment to the code. I have resubmit it as a normal comment. My apologies.

> * Coincidentally, I managed to get 100 % test coverage on `_fix_inout_args()`.

I was able to confirm this in my environment. Thank you.

> * What I also found was that running the tests on the original, bugged code in `_memberspec.py` crashes the python interpreter silently.

I am also not sure what happens when the AppVeyor encounters this situation.
This is the first time when we try to cover all `_fix_inouit_args` conditional branches in the tests. And perhaps we have only used it in a way that it works by chance so far.

> * While looking for real-life test subjects I came across a function with a parameter that has neither `'out'` nor `'in'` specified. I will investigate and open a separate issue.

Please do so.

Looking at the comments around the time `_fix_inout_args` was added, the developers might have expected solving these problems on `ctypes` side.
([Thomas Heller, the originator of this package, was a maintainer of `ctypes`](https://blog.python.org/2011/04/thomas-heller-steps-down-as-ctypes.html).)

https://github.com/enthought/comtypes/blob/cc9a0131edc76bd92073f75e9737aad40cd10c58/comtypes/__init__.py#L588-L597

In some cases, we might call out to the cpython core developers, as in #212.

jonschz:
Thank you for your work and input! If I may make a suggestion:
- We wait for a comment by @vasily-v-ryabov and then (presumably) merge the minimal impact fix of `_memberspec.py` into `drop_py2`, together with a collection of passing (and possibly xfailing) tests.
- I create a separate backport pull request for the `master` branch containing only the bugfix and no new tests.
- We discuss the unexpected behaviour of `_fix_inout_args()` in a separate issue.
- I don't think the python crashes need to be discussed here (as you said, this is a `cpython` issue), I was merely wondering if there's a chance the CI here doesn't handle a situation like this correctly.

junkmd:
Certainly Python 2.7 does not have `unittest.mock` support, we will have to do it without testing  if we backport this to `master`.

Anyway, let's wait for response from @vasily-v-ryabov on what to do about the bugs that be in `rescode` and `outargs`.

junkmd:
@vasily-v-ryabov 
Any concerns on this?
I would appreciate your reply to this ping.

jonschz:
In anticipation to your responses I have updated the code to be more in line with the legacy behaviour: If neither `'in'` nor `'out'` is specified, the argument will be treated as `'in'`, which is what the legacy code did. @junkmd, it'd be great if you could take a brief look.

@vasily-v-ryabov, here is a brief summary in case you do not want to read the entire thread:
- The bug I originally wanted to fix is fixed.
- The new code will behave like the old code on arguments that were not affected by this bug (with one minor difference in an edge case: the old code would crash on some invalid arguments, the new code throws an exception with a proper error message).
- We have added comprehensive unit tests for `_fix_inout_args`, mostly using `unittest.mock`. Testing all edge cases on real-world  examples proved to be impractical.
- We have found some odd behaviour in the original code. It is likely a minor oversight/slip by the original author and is **not** known to cause any issues; presumably, it has no impact at all. We have not modified this part and wrote tests to verify the current behaviour (just in case it changing it *would* have an effect we are not expecting).

junkmd:
Thank you for further research.

> If neither `'in'` nor `'out'` is specified, the argument will be treated as `'in'`, which is what the legacy code did

I think certainly because [a real COM library like `MSVidCtlLib` has such an implementation](https://github.com/enthought/comtypes/pull/473#issuecomment-1420706566).

My question is, can you actually instantiate the `IMFAttributes` defined in `MSVidCtlLib` like [`test_ienum`](https://github.com/enthought/comtypes/blob/532c3996e73866009249e02981d3ead9ae58d474/comtypes/test/test_ienum.py) and call `GetItemType` or `GetString`?

I know from previous discussions that it is difficult to test everything with real things. However, if it is possible to do even part of it with real things, I think it is better to do it.

jonschz:
I have added a real-life test as requested. @junkmd , could you clear the "Changes requested" review?

junkmd:
Thank you for all your efforts.
The test you added is also a good use case for `windll.mfplat.MFCreateAttributes`.

I added the `comtypes.client.GetModule("msvidctl.dll")` runtime warning suppression that was in `test_client` and `test_ienum`.

I also confirmed that CIs are passed.


<details>
<summary>memo</summary>

https://ci.appveyor.com/project/cfarrow/comtypes/builds/46670688
https://github.com/enthought/comtypes/actions/runs/4579861707/jobs/8088032533?pr=473
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/46670707

</details>

Let's wait for @vasily-v-ryabov's review.

comtypes/_memberspec.py
```diff
@@ -173,62 +173,97 @@ def _fix_inout_args(
...
+        #   instead of
+        #
+        #   rescode = rescode.__ctypes_from_outparam__()
+        #
+        #
+
         if outnum == 1:  # rescode is not iterable
             if len(outargs) == 1:
                 rescode = rescode.__ctypes_from_outparam__()
+                # rescode = outargs[0].__ctypes_from_outparam__()
```
vasily-v-ryabov, line 266: I think this comment is not necessary since such line is mentioned in the above large comment.

comtypes/_memberspec.py
```diff
@@ -173,62 +173,97 @@ def _fix_inout_args(
...
+            if dir_out:
                 outnum += 1
+            if dir_in:
+                param_index += 1
+
         rescode = func(self, *args, **kw)
         # If there is only a single output value, then do not expect it to
         # be iterable.
+
+        # My interpretation of this code:
```
vasily-v-ryabov, line 243: It would be nice to post your GitHub nickname here to let others understand who left this comment without looking into Git history.

comtypes/_memberspec.py
```diff
@@ -173,62 +173,97 @@ def _fix_inout_args(
...
+        # - 'outnum' counts the total number of 'out' and 'inout' arguments.
+        # - Confusingly, 'outargs' is a dict consisting of the supplied _inout_ arguments.
+        # - The call to func() returns the 'out' and 'inout' arguments.
+        #   Furthermore, it changes the variables in 'outargs' as a "side effect"
+        # - In a perfect world, it should be fine to just return 'rescode', but I assume there is a reason
+        #   why the original authors did not do that. Instead, they replace the inout variables in 'rescode'
+        #   by those in 'outargs', and call __ctypes_from_outparam__() on them.
+        #
+        # - What I don't understand is: Why is the behaviour different for outnum==1?
```
vasily-v-ryabov, line 252: I think the only reason is that singe `rescode` is not a tuple (iterable) so it's not possible to convert it to list by `list(rescode)`. It can be easily rewritten by creating list with `rescode = [rescode]`, but there is no such need as current implementation should be very little faster. :)

comtypes/_memberspec.py
```diff
@@ -173,62 +173,97 @@ def _fix_inout_args(
...
                     kw[name] = v
-            elif direction & 2 == 2:
+                else:
+                    # no parameter was passed, make an empty one of the required type
+                    # and pass it as a keyword argument
+                    v = atyp()
+                    if name is not None:
+                        kw[name] = v
+                    else:
+                        raise Exception("Unnamed inout parameters cannot be omitted")
```
vasily-v-ryabov, line 232: I'm not sure if it is more properly to raise `COMError` or maybe `TypeError` which seems used pretty often in similar cases.

After quick search across the whole repository I could find even such line:
```python
raise RuntimeError("BUG")
```
🤣

vasily-v-ryabov (approved):
In general the logic looks OK to me. I can only rely on your testing. So please proceed with the fix.

And sorry for such a long waiting. Hope @junkmd can finish work on release `1.3.0`. Please don't wait for my opinion if I don't reply for more than 1 week. Most probably 2024 will be too busy again for me. 😄 But I never say "never" to open source projects! 😄

comtypes/_memberspec.py
junkmd: I think it would be more appropriate to raise a `TypeError` here.

When throwing a `COMError`, I believe it should be raised as if it were thrown by the COM system of the execution environment (via CPython's `_ctypes`).
We can instantiate `_ctypes.COMError` by passing the error code (`hresult`), message, and a 5-length tuple as constructor arguments and then raise it, I am not sure about the correct combination of arguments in this case.
It seems to be a challenging decision within the community.

Instead, considering the situation as "unable to handle properly as a Python method type," I suggest raising a `TypeError`.

junkmd: > After quick search across the whole repository I could find even such line:
> 
> ```python
> raise RuntimeError("BUG")
> ```
> 
> 🤣

Oh, this is indeed one of those legacies that hasn't changed since this package was created.
In modern Python, carefully scrutinizing the types of errors thrown and crafting elegant messages can make the codebase more robust, effectively communicating the intentions of the implementer to future maintainers.

comtypes/_memberspec.py
junkmd: I agree with this point.

It is necessary to add comments explaining why `outargs[0]` and `rescode` indicate the same thing when `outnum == 1` and `len(outargs) == 1`.
- It would also be useful to comment on the line where the variable declaration is located.

However, regarding the implementation, I believe it's fine as it is, considering the desire to minimize changes and the slightly faster processing speed.

junkmd:
@jonschz 

I appreciate your waiting.
Are you still interested in this topic?

jonschz:
Quick with your responses as always, I appreciate it! I'll most likely update the branch today or tomorrow.

comtypes/_memberspec.py
jonschz: > It would also be useful to comment on the line where the variable declaration is located.

What exactly would you like to see commented there, that is not redundant with the information in the big comment block? If you have a specific idea, feel free to change it and then merge. Thank you for your efforts!

vasily-v-ryabov: In my brief understanding the comment lines 252 - 260 are not necessary. Though some comment about better speed is also OK.

junkmd: > > It would also be useful to comment on the line where the variable declaration is located.
> 
> What exactly would you like to see commented there, that is not redundant with the information in the big comment block? If you have a specific idea, feel free to change it and then merge. Thank you for your efforts!

I tried moving the comment to the variable declaration section, but it became confusing.

I am currently considering comments that would be helpful to future maintainers.
I will add a commit with corrected comments after this.

junkmd:
I added a commit.

I updated the method and class name quoting in some comments.
For `outnum==1`, only processing speed was mentioned because I rethink that it would be better to read the codebase than quite long comments to understand the implementation.

comtypes/test/test_inout_args.py
```diff
@@ -0,0 +1,490 @@
...
+        self_ = MagicMock(name="Self")
+        spec = comtypes.COMMETHOD(
+            [],
+            HRESULT,
+            "Foo",
+            (["in", "out"], POINTER(c_ulong)),
+        )
+        orig = MagicMock()
+        fixed = _fix_inout_args(orig, spec.argtypes, spec.paramflags)
+        with self.assertRaises(Exception) as cm:
```
junkmd, line 482:
```suggestion
        with self.assertRaises(TypeError) as cm:
```
The error types in the production code have changed, and the tests should be changed accordingly.

junkmd (approved):
Thank you for dealing with this topic over a long period of time.

This was a very old bug, as the PR title suggests, as well as production code that had not been boundary value tested before.

Your contribution is very beneficial.

jonschz:
Thank you for your efforts at maintaining this package! You greatly contributed to the end result here.

PR #484, opened 2023-06-04
change `_MemberSpec`s to `NamedTuple`s

junkmd:
#392 

In versions where `typing` is supported, it should be more elegant to use `NamedTuple` than to reproduce the `__iter__` behavior of the tuple in a custom class.

The `NamedTuple` has a [restriction on inheritance](https://github.com/python/cpython/blob/ce558e69d4087dd3653207de78345fbb8a2c7835/Lib/typing.py#L2670-L2675), so `_MemberSpec` cannot be used as the base class as before.

So the current `_MemberSpec` is a Union type of `_ComMemberSpec` and `_DispMemberSpec`.

### DIFF
--- a/comtypes/__init__.py
+++ b/comtypes/__init__.py
@@ -744,9 +744,7 @@ def DISPPROPERTY(idlflags, proptype, name) -> _DispMemberSpec:
 #     )
 
 
-def COMMETHOD(
-    idlflags, restype, methodname, *argspec
-) -> _ComMemberSpec:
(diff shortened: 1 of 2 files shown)


junkmd:
CIs are passed.

<details>
<summary>memo</summary>

https://ci.appveyor.com/project/cfarrow/comtypes/builds/47216094
https://github.com/enthought/comtypes/actions/runs/5167462859/jobs/9308358805?pr=484
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/47216068

</details>

Issue #400, opened 2022-12-11
Add inline type annotations to dynamically generated modules


junkmd:
In #327, we were planning to implement a feature to generate a `.pyi` file with its type hints as soon as the module is generated by `client.GetModule`.

However, with a suggestion from @kdschlosser, we shifted to generating modules with inline type annotations.

Inline type annotations are only available in Python 3.x and later, so the merge destination for the contributions is **[`drop_py2` branch](https://github.com/enthought/comtypes/tree/drop_py2)**.

As planned in #392, we will support Python3.7 and later versions, so type hinting should be done in the way supported from 3.7 and earlier.  
For example...
- [`Literal`  (new in Python3.8)](https://docs.python.org/3/library/typing.html#typing.Literal), [`Final` (new in Python3.8)](https://docs.python.org/3/library/typing.html#typing.Final), [`Self` (new in Python3.10)](https://docs.python.org/3/library/typing.html#typing.Self), and other new symbols are not available.
- [PEP585](https://peps.python.org/pep-0585/) is not available.
- [PEP604](https://peps.python.org/pep-0604/) is not available.

This kanban will be updated as the situation requires.

Any opinions would be appreciated.

Issue #371, opened 2022-11-07
support for named arguments and default argument values


junkmd:
'named arguments' feature is 'should be added' since 16 years ago, according to commented(938fbd50d2077c6bd524790a9ed798c3990cdfc0) by Thomas Heller, the originator of this package.

https://github.com/enthought/comtypes/blob/dbe4fa1bed94fef20b9c56faad1693230dd32c43/comtypes/__init__.py#L609-L611

Also, `bound_named_property.__call__` should support them.

https://github.com/enthought/comtypes/blob/dbe4fa1bed94fef20b9c56faad1693230dd32c43/comtypes/__init__.py#L897-L917

If functionality is added, this code snippet will work!

```py
xl = CreateObject("Excel.Application")
wb = xl.Workbooks.Add()
assert xl.Range["A1:C3"].Address() == "$A$1:$C$3"   # it is possible before the functionality is added
assert xl.Range["A1:C3"].Address(None, None) == "$A$1:$C$3"  # it is possible before the functionality is added
assert xl.Range["A1:C3"].Address(ColumnAbsolute=False) == "A$1:C$3"  # not wroking currently, I want THIS
assert xl.Range["A1:C3"].Address(None, False) == "A$1:C$3"   # it is possible before the functionality is added
assert xl.Range["A1:C3"].Address(None, ColumnAbsolute=False) == "A$1:C$3"  # not wroking currently, I want THIS
assert xl.Range["A1:C3"].Address(RowAbsolute=False) == "$A1:$C3"  # not wroking currently, I want THIS
assert xl.Range["A1:C3"].Address(ReferenceStyle=xlR1C1) == "R1C1:R3C3"  # not wroking currently, I want THIS
```

`_DispMemberSpec` has `argspec`, so I think it may be easy to get names and default values.

PR #475, opened 2023-01-24
Add enumeration definitions in the friendly modules

junkmd:
#345

I have modified the `codegenerator` so that the Python friendly enumerations are defined in the friendly module as derived classes of `enum.IntFlag`.

Specifically, for example, the `Scripting.py` changes as follows before and after this change.

```py
import comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0 as __wrapper_module__
from comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0 import (
    FileAttribute, DriveTypeConst, TristateMixed, IFileSystem,
    SystemFolder, IFileCollection, StandardStreamTypes,
    _check_version, GUID, IFolderCollection, Dictionary, Directory,
    Remote, Library, StdOut, IOMode,
    __MIDL___MIDL_itf_scrrun_0001_0001_0003, WindowsFolder, VARIANT,
    Alias, helpstring, TristateUseDefault, IScriptEncoder, Archive,
    TextStream, IDrive, IFile, HRESULT, Hidden, StdErr, Folders,
    dispid, typelib_path, IFileSystem3, ReadOnly, ForWriting, StdIn,
    FileSystemObject, ForAppending, Encoder, CDRom, IDriveCollection,
    COMMETHOD, Volume, CompareMethod, BinaryCompare, IDictionary,
    BSTR, Normal, TristateTrue, TristateFalse, Fixed, UnknownType,
    Files, File, _lcid, Tristate, SpecialFolderConst, ForReading,
    ITextStream, __MIDL___MIDL_itf_scrrun_0001_0001_0001, Removable,
    RamDisk, VARIANT_BOOL, Drives, Folder, TemporaryFolder, CoClass,
    __MIDL___MIDL_itf_scrrun_0000_0000_0001, System,
    __MIDL___MIDL_itf_scrrun_0001_0001_0002, Compressed, TextCompare,
    DatabaseCompare, IFolder, IUnknown, Drive
)


__all__ = [
    'Encoder', 'FileAttribute', 'CDRom', 'IDriveCollection',
    'DriveTypeConst', 'TristateMixed', 'Volume', 'SystemFolder',
    'IFileSystem', 'IFileCollection', 'CompareMethod',
    'StandardStreamTypes', 'BinaryCompare', 'IDictionary', 'Normal',
    'TristateTrue', 'TristateFalse', 'Fixed', 'IFolderCollection',
[...]

PR #486, opened 2023-06-10
Support named args for dispmethod

junkmd:
#371 

I was trying to implement the #400 functionality.
However, I realized that it is misleading to have a function annotation that allows named arguments to be accepted, since a dispmethod called via `IDispatch.Invoke` will raise an error when a named (with default value) argument is passed at runtime.

So, I improved the process of instantiating `DISPPARAMS` so that it can also accept named arguments in dispmethod.
Using `_argspec` passed to `IDispatch.Invoke`, it completes default values for optional arguments that were not passed, to ensure that the passed argument names are correct.

This added functionality is only triggered when named arguments are passed to dispmethod.
Same as before, when just only positional arguments are passed, the VARIANT array is assigned to `DISPPARAMS.rgvarg` in the reverse order.
Therefore, this change does not break backward compatibility.
In the case that an element of `argspec` is not what the newly added processing expects, and an error is raised when passing named arguments, it can be avoided by passing only positional arguments.

The `bound_named_property` has also been improved to allow passing named arguments.

As for testing, I used Excel to verify that it is possible to pass named arguments to a dispmethod.
I have Excel in my development environment, so I can run this test and make sure it passes.
However, AppVeyor CI does not have Excel, so I am not able to fully guarantee End2End operation with CI.
Instead, I generously cover the test of `DISPPARAMS` instantiation process, which can be executed independent of Excel and the environment.
commethod (and stdmethod) were originally designed to accept named arguments, but there were no tests to do so explicitly, so I added them.

If there is a dispmethod in an environment-independent library that can easily provide fixtures, I will try to test with that object as well, so I am looking for feedback.

### DIFF
--- a/comtypes/_memberspec.py
+++ b/comtypes/_memberspec.py
@@ -445,12 +445,11 @@ class DispMemberGenerator(object):
 
     # Should the funcs/mths we create have restype and/or argtypes attributes?
     def _make_disp_method(self, m: _DispMemberSpec) -> Callable[..., Any]:
-        memid = m.memid
         if "propget" in m.idlflags:
 
             def getfunc(obj, *args, **kw):
                 return obj.Invoke(
-                    memid, _invkind=2, *args, **kw
+                    m.memid, _invkind=2, _argspec=m.argspec, *args, **kw
                 )  # DISPATCH_PROPERTYGET
 
             return getfunc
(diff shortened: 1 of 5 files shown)


junkmd:
CIs are passed.

<details>
<summary>memo</summary>

https://ci.appveyor.com/project/cfarrow/comtypes/builds/47271940
https://github.com/enthought/comtypes/actions/runs/5230311902/jobs/9443793826?pr=486
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/47271962

</details>

junkmd:
@vasily-v-ryabov 

Please review this.

Not to break backwards compatibility, but since this is more of a "feature addition to allow named arguments to be passed" than a "bug fix for not being able to pass named arguments", and since there are many lines of codebase changed.

If we decide not to incorporate this into the `1.3.0` like #475, I would move #400 and this from the [`1.3.0` milestone](https://github.com/enthought/comtypes/milestone/4) to next version milestone.

PR #523, opened 2024-04-07
Bump version and update change log for 1.4.0

junkmd:
1.4.0 is the first major version supporting enumerations in friendly modules.

[The names that were used as aliases for ``c_int`` within friendly modules will now be used as the names of enumeration types](https://github.com/enthought/comtypes/blob/0d43d183e1881ab76bacf64d370c1697e21963b7/docs/source/index.rst?plain=1#L668-L691).


junkmd:
One of the tests that was not skipped in AppVeyor was skipped in GitHub Actions.

```
test_relpath (test_client.Test_GetModule) ... skipped 'This depends on typelib and test module are in same drive'
```

https://github.com/enthought/comtypes/blob/0d43d183e1881ab76bacf64d370c1697e21963b7/comtypes/test/test_client.py#L24-L32

This is because `scrrun.dll`, which was on the C drive in the AppVeyor environment, is on a different drive in the GHA environment.
In my local environment, both `comtypes` and `scrrun.dll` are on the C drive, so the test is not skipped and passes.

Issue #524, opened 2024-04-09
'SyntaxError: invalid syntax' occurs during MS Project module generation with GetModule function 


kateryna-ruzhytska:
MS_PROJECT_HASH = '{A7107640-94DF-1068-855E-00DD01075445}'
com.GetModule((MS_PROJECT_HASH, 4, 0)) returns 'SyntaxError: invalid syntax' starting from 1.4.0 version:

`File "C:\Python39\lib\site-packages\comtypes\gen\MSHTML.py", line 4
    from comtypes.gen._3050F1C5_98B5_11CF_BB82_00AA00BDCE0B_0_4_0 import 
                                                                                                                                    ^
SyntaxError: invalid syntax`

![comtypes_syntax_error_1 4 0](https://github.com/enthought/comtypes/assets/8039690/f074e70e-963d-4308-a1c3-9a49c5625662)

Current issue is reproducible periodically.
Sometimes it happens to import appropriate classes but is it expected to have the second import?

![comtypes_1 4 0](https://github.com/enthought/comtypes/assets/8039690/dcd90e12-4d37-42b2-a53a-9d87dfcd83af)

junkmd:
> Sometimes it happens to import appropriate classes but is it expected to have the second import?

Yes, indeed, it is expected behavior that classes are imported in "the second import".

kateryna-ruzhytska:
Can you please try to reproduce this issue by the steps below.

Precondition: comtypes version is 1.4.0

**Steps to reproduce:**
1. Remove folder <your_python_path>\Lib\site-packages\comtypes\gen
2. Run command in Command Prompt: 
`python -c "import comtypes.client as com; com.GetModule('C:\Windows\System32\mshtml.tlb')"`

**NOTE: you may need to perform the steps above several times to reproduce.**

Actual result:

> Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\Kateryna\.pyenv\pyenv-win\versions\3.9.4\lib\site-packages\comtypes\client\_generate.py", line 124, in GetModule
    return ModuleGenerator().generate(tlib, pathname)
  File "C:\Users\Kateryna\.pyenv\pyenv-win\versions\3.9.4\lib\site-packages\comtypes\client\_generate.py", line 203, in generate
    return self._create_friendly_module(tlib, modulename)
[...]

PR #527, opened 2024-04-15
Define constants and `c_int` aliases earlier than others in wrapper modules

junkmd:
I have noticed something while fixing the bug reported in #524.

In the wrapper module, members of an enumeration are defined as module-level constants.
If they have the same name as a previously defined interface or CoClass, it will overwrite the definition.

The order of type information parsed from the COM library is optimized for the environment, so it is not always fixed.
In other words, symbols with duplicate names cannot be identified as to what type or value they are until the module is generated.

To ensure that the values of constants do not overwrite interfaces or CoClasses, codegenerator generates the codebase that defines the constants before them.


junkmd:
CIs are passed.
https://github.com/enthought/comtypes/actions/runs/8687275606
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/49616197

PR #529, opened 2024-04-15
Determine whether a COM interface is one of the known symbols not only by its name but also by using its iid

junkmd:
While fixing the bug reported in #524, I noticed something.

If there is an interface in the COM type library with the same name as a symbol defined in `ctypes` or `ctypes.wintypes`, the `codegenerator` can only identify the symbol by its name, which leads to the symbol being imported and used inappropriately.

To prevent this confusion, the `codegenerator` will now determine whether a COM interface is one of the known symbols not only by its name but also by using its iid.

### DIFF
--- a/comtypes/client/_generate.py
+++ b/comtypes/client/_generate.py
@@ -1,11 +1,12 @@
 from __future__ import print_function
 import ctypes
 import importlib
+import inspect
 import logging
 import os
 import sys
 import types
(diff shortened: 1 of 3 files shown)


junkmd:
CIs are passed.

https://github.com/enthought/comtypes/actions/runs/8697636479
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/49622407

PR #531, opened 2024-04-22
Removing trailing newlines generated by `codegenerator` and `typeannotator`

junkmd:
`ComInterfaceMembersAnnotator`, `DispInterfaceMembersAnnotator`, `ComMethodGenerator`, `DispMethodGenerator`, and `DispPropertyGenerator` had a newline at the end of the codebase returned from `generate`.

When adding these codebases to wrapper modules, `end=""` was passed to `print` to avoid generating surplus newlines.
However, the inconsistency in the use of print could potentially increase cognitive load during future development.

By standardizing the use of `print`, we can reduce maintenance costs in the future.

### DIFF
--- a/comtypes/tools/codegenerator.py
+++ b/comtypes/tools/codegenerator.py
@@ -1149,11 +1149,11 @@ class CodeGenerator(object):
             print("            return item", file=self.stream)
             print("        raise IndexError(index)", file=self.stream)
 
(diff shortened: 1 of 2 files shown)


junkmd:
CIs are passed.

https://github.com/enthought/comtypes/actions/runs/8784745131
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/49665846

PR #532, opened 2024-04-22
add `Incomplete` and `Hresult` to `hints.pyi`

junkmd:
## `Hresult`
The `HRESULT` will be returned by calling a com method with no `[out]` or `[out, retval]` arguments.

To replicate this runtime behavior, such methods will now be annotated with an integer type for their return value.

However, leaving it as `int` could lead to misinterpretation of the value, so a type alias named `Hresult` is prepared to represent that it is one of the integer values of HRESULT.

## Incomplete`
Parts that cannot be annotated with a detailed type by the `typeannotator` will be annotated as `Incomplete`.
This name was inspired by `typeshed`.

### DIFF
--- a/comtypes/hints.pyi
+++ b/comtypes/hints.pyi
@@ -22,9 +22,9 @@ if sys.version_info >= (3, 8):
 else:
     from typing_extensions import Protocol
 if sys.version_info >= (3, 10):
-    from typing import Concatenate, ParamSpec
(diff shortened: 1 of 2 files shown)


junkmd:
CIs are passed.

https://ci.appveyor.com/project/pywinauto/pywinauto/builds/49669170

https://github.com/enthought/comtypes/actions/runs/8792259118

PR #533, opened 2024-04-25
Prevent the occurrence of `SyntaxError` in friendly modules.

junkmd:
#524 

See also https://github.com/enthought/comtypes/issues/524#issuecomment-2049546592 and https://github.com/enthought/comtypes/issues/524#issuecomment-2071955502.

I will wait for feedback from the community, and if there are no objections, I will merge this and release it as `comtypes==1.4.2` in early May.

### DIFF
--- a/comtypes/client/_generate.py
+++ b/comtypes/client/_generate.py
@@ -1,11 +1,12 @@
 from __future__ import print_function
 import ctypes
 import importlib
+import inspect
 import logging
 import os
 import sys
 import types
-from typing import Any, Tuple, List, Optional, Dict, Union as _UnionT
+from typing import Any, Tuple, List, Mapping, Optional, Dict, Union as _UnionT
 import winreg
 
 from comtypes import GUID, typeinfo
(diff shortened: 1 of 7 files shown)


junkmd:
CIs are passed.

https://ci.appveyor.com/project/pywinauto/pywinauto/builds/49693478
https://github.com/enthought/comtypes/actions/runs/8840682732

junkmd:
I have added the test to verify that `GetModule` regenerates the module.
- In the codebase prior to this PR, `test_friendly_module_is_missing` would result in a `SyntaxError` similar to #524.
- Modules generated from `mshtml.tlb` are large in size and regenerating them multiple times makes the test time-consuming. Since any COM type library can be used to reproduce the situation where the existing module file does not exist, I am using the smaller `scrrun.dll` for testing.

This is similar to the existing `test_findgendir.py` in that it patches `sys.modules`.

This test involves a complex interplay of context-managers, `sys.modules`, `sys.path`, `importlib`, and `mock.patch`.
Consequently, the codebase has become complex.

To clarify the behavior within the fixtures, I've inserted `assert` statements and comments at key points.
These aims to facilitate asynchronous communication with future maintainers.

I welcome any feedbacks as well.

junkmd:
> the `generate` method to not generate a module

If you mean that `ModuleGenerator.generate` is not adhering to the single responsibility principle because it is not only generating new modules but also responsible for loading existing modules, then I agree.

Therefore, I have made changes so that the attempt to load existing modules is done within `GetModule`, before `ModuleGenerator.generate` is called.

cfarrow:
Thank you for explaining this to me. My understanding of the code was not complete. I agree that the new responsibility belongs in `GetModule`. The test is a good addition with the new behavior.

junkmd:
Thank you for your reply.

When I first looked at the codebase for `comtypes==1.1.11`, it took me some time to understand this behavior because my initial expectations and the actual implementation were different.

Thank you.

PR #534, opened 2024-04-30
Determine whether a disp interface is one of the known symbols not only by its name but also by using its iid

junkmd:
As in https://github.com/enthought/comtypes/pull/529, I have also implemented the same thing for disp interfaces, not only com interfaces.


junkmd:
CIs are passed.

https://ci.appveyor.com/project/pywinauto/pywinauto/builds/49720843
https://github.com/enthought/comtypes/actions/runs/8894080033

PR #539, opened 2024-05-10
Add a feature to `typeannotator` that interprets the return value

junkmd:
I have made it so that the return type of each method is statically typed based on type library information.

There are still many parts that are `Incomplete`, but I will add more features in the future.


junkmd:
CIs are passed.

https://github.com/enthought/comtypes/actions/runs/9030956183
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/49788060

PR #540, opened 2024-05-11
update type annotations for `_IUnknown_Base` and `IUnknown`

junkmd:
Defining `__com_AddRef`, `__com_Release`, and `__com_QueryInterface` as members of `_IUnknown_Base`, type checkers interpret these members as accessible not only to `IUnknown` but also to their subclasses.

I think that this could lead to confusion, I choose to remove the definition of these members from `_IUnknown_Base`.


junkmd:
CIs are passed.

https://ci.appveyor.com/project/pywinauto/pywinauto/builds/49792592
https://github.com/enthought/comtypes/actions/runs/9039394519

Issue #500, opened 2023-12-14
Got a syntax error  when ran on Win Server 2019 standard


mittu80:
Hello, 

I tried to import module  pywifi and its dependancy comtypes but got the following syntax error.
__init__.py line 381
![image](https://github.com/enthought/comtypes/assets/84013619/23010a8d-0831-4d93-9f4f-eff05f4c3e64)

Can you help me fix this?

junkmd:
Hi,

I think that the `comtypes` in your environment is a quite old version from the time when it supported Python 2.6.
By executing `pip install -U comtypes`, you can update to the current version that natively supports Python 3.x.

mittu80:
Hi, 
Thanks for your answer. I've tried the latest 1.2.0 as well unfortunately i received a a ctypes DLL Load error.
I am suspecting the latest rel is not compatible to windows server (virtual pc) as when i ran on windows laptop it worked.

junkmd:
Hi,

> I might be wrong here as in mycase its win server and not win CE.

I think that's probably the case.
One of the virtual environments I use for my main job is Windows Server, and `comtypes` is functioning on it.

[...]

PR #554, opened 2024-06-06
Stop mentioning the last version supporting Windows CE

forderud:
Follow-up to #211 that dropped Windows CE support back in 2020. I doubt that any new comtypes user is interested in figuring out the last version that supported Windows CE these days. Therefore suggest removing the sentence from the front page since it's of low relevance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@
 `comtypes` allows to define, call, and implement custom and dispatch-based COM interfaces in pure `Python`.
 
 This package works on Windows only.
-- [`comtypes==1.1.7`](https://pypi.org/project/comtypes/1.1.7/) is the last version supporting Windows CE.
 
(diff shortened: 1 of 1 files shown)


junkmd (approved):
Thank you.

Given the discussion in #500, it's true that mentioning Windows CE might rather invite confusion with Windows Server.

This package still contains workarounds that have become dead code, remnants from when it supported WinCE. If there are proposals from the community to remove these from the codebase, I would accept them,  same as this PR.

junkmd:
As an additional note, [support for WinCE also ended on October 10, 2023](https://learn.microsoft.com/en-us/lifecycle/end-of-support/end-of-support-2023#products-reaching-end-of-support).
https://learn.microsoft.com/en-us/previous-versions/windows/iot-core/windows_ce/windows-ce-migration-faq

Issue #216, opened 2020-07-23
Future of this package


lassoan:
This package is very useful, thanks a lot for providing it! I would like to use it in a large open-source project ([3D Slicer](https://github.com/Slicer/Slicer)) but I'm not sure if I can rely on it in the long term, due to the followings:

1. It seems that there have been no new tags (and releases on PyPI) for almost two years
2. It is still not a native Python3, but patched using 2to3
3. Large backlog of open issues and 10+ open pull requests

Could somebody provide information about current status and future plans for this package?

I see that there have been some recent commits, so I still have some hope. Thanks in advance!

cfarrow:
Hi Andras. I know of no long term plans for this package, except perhaps for users keeping it running for their use cases. The package can live on for a while in this self-service mode, but nice-to-haves like native python 3 support are not likely be prioritized.

I do virtually no work on this package, as my job and personal life no longer involve coding. @vasily-v-ryabov does the bulk of reviewing and committing. I'd welcome a discussion about building a critical mass of contributors and admins for this project.

lassoan:
@cfarrow thanks a lot for the quick answer. @vasily-v-ryabov do you have any comments?

vasily-v-ryabov:
Hi guys, I have some plans on diving deeper into the comtypes source code with potential improvements. My interest is mostly comtypes' usage in pywinauto as a dependency (we're already using it for few years). I'll try to explain my immediate interests and nice-to-have things which I can support as a reviewer.

 * I'm not pursuing better Python 2.x/3.x support, but can review/discuss it.
 * Running all unit tests on AppVeyor would be more helpful for potential contributors. It is also nice to have.
 * The most critical for now is new wheel versions support which is reported in both projects already. Hope to take this task in nearest two months if nobody takes care of it. It could be a reason for 1.1.8 version update.
 * I do want to debug some issues with 32-bit/64-bit Python / COM libraries that we have in pywinauto's backlog. This is medium priority for me, but there are a lot of high priority tasks in pywinauto which is also a hobby project in its turn. :)
 * Also I'm interested in more convenient packaging into a single executable (low priority for now, may be raised in the future) so I may think about on-the-fly code generation into a memory without saving to cache file or something like that.

No more global things for now.

lassoan:
A new release would be important, because if there are no new releases then fixes and improvements will not get to users. It would be also important to merge pull request and review issues. Setting up AppVeyor would be nice because then pull requests could be more confidently merged.

[...]

PR #556, opened 2024-06-07
Declare the minimum Python version as 3.7.

jaraco:
Ref https://github.com/enthought/comtypes/issues/216#issuecomment-2154948863

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[options]
+python_requires = >=3.7


junkmd (approved):
Thank you for your quick response.

I'm considering releasing `comtypes==1.4.4` after merging this. Is that okay?

junkmd:
I have looked at #557.

PR #558, opened 2024-06-07
Bump version and update change log for 1.4.4

junkmd:


junkmd:
Thanks @jaraco and @forderud!

PR #569, opened 2024-06-11
Implement SAFEARRAY pointers and SAFEARRAYs as method parameters of a Dispatch Interface

geppi:
With a COM Dual Interface it is possible to get server side changes to SAFEARRAYs marshaled back to the client.
However, for a pure IDispatch interface this is currently not implemented.

This pull request adds handling of SAFEARRAYs to the implementation of the automation interface.

A unittest and a new COM test interface for the C++ COM test server are also added.

I have to apologize for a mistake I made in [#535](https://github.com/enthought/comtypes/pull/535).
Back at the time I did favor the implementation of several related interfaces in a single component.
Unfortunately I didn't realize that this is a problem when the interfaces inherit from the same interface but require different implementations for this interface as is the case for IDispatch (see [Multiple Dual Interfaces](https://learn.microsoft.com/en-us/cpp/atl/multiple-dual-interfaces)).

While adding another component to the C++ COM test server was not a big deal, the component naming scheme for which I had advocated in [#535](https://github.com/enthought/comtypes/pull/535) is inadequate for this setup.

So I was wrong and therefore have now also changed the name of the component that implements the interface for record parameter testing to reflect its limited scope.

In the future, adding additional interfaces and components implementing them will be more straight forward.

PR #575, opened 2024-06-23
Name change of component implementing record parameter testing.

geppi:
Background: Implementing several related interfaces in a single component is a problem when the interfaces inherit from the same interface but require different implementations for this interface as is the case for IDispatch. Therefore additional interfaces will be implemented in separate components.

The name change reflects the limited scope of the component.


junkmd (approved):
Thank you.
Please rebase #569.

PR #576, opened 2024-06-25
Change the default value of the 'extra' parameter in the 'create' method of tagSAFEARRAY subtypes.

geppi:
With the current setting of [extra=None](https://github.com/enthought/comtypes/blob/11d7651e5ed04100ca0388e9617395cfc24092cf/comtypes/safearray.py#L112) you need to go through the following steps to create a SAFEARRAY of e.g. VT_RECORDs:
```
import comtypes
from comtypes.gen import SomeTypeLib

record_list = [SomeTypeLib.T_SOME_RECORD(), SomeTypeLib.T_SOME_RECORD()]
record_info = comtypes.typeinfo.GetRecordInfoFromGuids(*SomeTypeLib.T_SOME_RECORD._recordinfo_)
safearray_of_records = comtypes.safearray._midlSAFEARRAY(SomeTypeLib.T_SOME_RECORD).create(record_list , extra=record_info )
 ```
As the comment following [line 124 in safearray.py](https://github.com/enthought/comtypes/blob/11d7651e5ed04100ca0388e9617395cfc24092cf/comtypes/safearray.py#L124) says:
```
# For VT_UNKNOWN or VT_DISPATCH, extra must be a pointer to
# the GUID of the interface.
#
# For VT_RECORD, extra must be a pointer to an IRecordInfo
# describing the record.
 ```
Admittedly, some of the code in _safearray.py_ is deep Python black magic.
However, the _extra_ value required for the creation of SAFEARRAYs with VT_RECORD, VT_DISPATCH or VT_UNKNOWN elements is already identified at the time when the associated tagSAFEARRAY subtype is created in the [_make_safearray_type](https://github.com/enthought/comtypes/blob/11d7651e5ed04100ca0388e9617395cfc24092cf/comtypes/safearray.py#L81) method.

As far as I understand the code, the value of the local identifier _extra_ in __make_safearray_type_ is preserved in the closure of the _POINTER(sa_type)_. This can be seen in action e.g. in the [from_param ](https://github.com/enthought/comtypes/blob/11d7651e5ed04100ca0388e9617395cfc24092cf/comtypes/safearray.py#L212)method:
 ```
        @classmethod
        def from_param(cls, value):
            if not isinstance(value, cls):
                value = cls.create(value, extra)
                value._needsfree = True
            return value
```
To be quite frank, I don't understand why the _extra_ parameter in the _create_ method is required at all because the method has access to the same closure value. Maybe it's just to keep some flexibility or do I miss something?

Even less do I understand why the default setting is `None` because it requires the user code to perform the same procedure of getting a pointer to the IRecordInfo or the GUID of the VT_UNKNOWN or VT_DISPATCH interface again.

With the suggested change, the steps to create a SAFEARRAY of e.g. VT_RECORDs are reduced to:
```
import comtypes
from comtypes.gen import SomeTypeLib

record_list = [SomeTypeLib.T_SOME_RECORD(), SomeTypeLib.T_SOME_RECORD()]
safearray_of_records = comtypes.safearray._midlSAFEARRAY(SomeTypeLib.T_SOME_RECORD).create(record_list)
 ```
while the flexibility to pass a different value for the _extra_ parameter to the _create_ method is maintained.

### DIFF
--- a/comtypes/test/test_midl_safearray_create.py
+++ b/comtypes/test/test_midl_safearray_create.py
@@ -1,0 +1,92 @@
+# coding: utf-8
+
+from ctypes import byref, c_int, pointer, POINTER
+import unittest
+
+import comtypes
+import comtypes.safearray
+from comtypes import CLSCTX_INPROC_SERVER
+from comtypes.client import CreateObject, GetModule
+import comtypes.typeinfo
+
+GetModule("UIAutomationCore.dll")
+from comtypes.gen.UIAutomationClient import CUIAutomation, IUIAutomation
+
+GetModule("scrrun.dll")
+from comtypes.gen.Scripting import Dictionary, IDictionary
+
+ComtypesCppTestSrvLib_GUID = "{07D2AEE5-1DF8-4D2C-953A-554ADFD25F99}"
+
+try:
+    GetModule((ComtypesCppTestSrvLib_GUID, 1, 0, 0))
+    from comtypes.gen.ComtypesCppTestSrvLib import StructRecordParamTest
+
+    IMPORT_FAILED = False
+except (ImportError, OSError):
+    IMPORT_FAILED = True
+
+
+class Test_midlSAFEARRAY_create(unittest.TestCase):
+    def test_iunk(self):
+        extra = pointer(IUIAutomation._iid_)
+        iuia = CreateObject(
+            CUIAutomation().IPersist_GetClassID(),
+            interface=IUIAutomation,
+            clsctx=CLSCTX_INPROC_SERVER,
+        )
+        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IUIAutomation))
+        for ptn, sa in [
+            ("with extra", sa_type.create([iuia], extra=extra)),
+            ("without extra", sa_type.create([iuia])),
+        ]:
+            with self.subTest(ptn=ptn):
+                (unpacked,) = sa.unpack()
+                self.assertIsInstance(unpacked, POINTER(IUIAutomation))
+
+    def test_idisp(self):
+        extra = pointer(IDictionary._iid_)
+        idic = CreateObject(Dictionary, interface=IDictionary)
+        idic["foo"] = "bar"
+        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IDictionary))
+        for ptn, sa in [
+            ("with extra", sa_type.create([idic], extra=extra)),
+            ("without extra", sa_type.create([idic])),
+        ]:
+            with self.subTest(ptn=ptn):
+                (unpacked,) = sa.unpack()
+                self.assertIsInstance(unpacked, POINTER(IDictionary))
+                self.assertEqual(unpacked["foo"], "bar")
+
+    @unittest.skipIf(IMPORT_FAILED, "This depends on the out of process COM-server.")
+    def test_record(self):
+        extra = comtypes.typeinfo.GetRecordInfoFromGuids(
+            *StructRecordParamTest._recordinfo_
+        )
+        record = StructRecordParamTest()
+        record.answer = 42
+        sa_type = comtypes.safearray._midlSAFEARRAY(StructRecordParamTest)
+        for ptn, sa in [
+            ("with extra", sa_type.create([record], extra=extra)),
+            ("without extra", sa_type.create([record])),
+        ]:
+            with self.subTest(ptn=ptn):
+                (unpacked,) = sa.unpack()
+                self.assertIsInstance(unpacked, StructRecordParamTest)
+                self.assertEqual(unpacked.answer, 42)
+
+    def test_ctype(self):
+        extra = None
+        cdata = c_int(1)
+        sa_type = comtypes.safearray._midlSAFEARRAY(c_int)
+        for ptn, sa in [
+            ("with extra", sa_type.create([cdata], extra=extra)),
+            ("without extra", sa_type.create([cdata])),
+        ]:
+            with self.subTest(ptn=ptn):
+                (unpacked,) = sa.unpack()
+                self.assertIsInstance(unpacked, int)
+                self.assertEqual(unpacked, 1)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/comtypes/test/test_midl_safearray_create.py
+++ b/comtypes/test/test_midl_safearray_create.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 
-from ctypes import byref, c_int, pointer, POINTER
+from ctypes import c_int, pointer, POINTER
 import unittest
 
 import comtypes
@@ -64,7 +64,9 @@ class Test_midlSAFEARRAY_create(unittest.TestCase):
             *StructRecordParamTest._recordinfo_
         )
         record = StructRecordParamTest()
+        record.question = "The meaning of life, the universe and everything?"
         record.answer = 42
+        record.needs_clarification = True
         sa_type = comtypes.safearray._midlSAFEARRAY(StructRecordParamTest)
         for ptn, sa in [
             ("with extra", sa_type.create([record], extra=extra)),
@@ -73,7 +75,12 @@ class Test_midlSAFEARRAY_create(unittest.TestCase):
             with self.subTest(ptn=ptn):
                 (unpacked,) = sa.unpack()
                 self.assertIsInstance(unpacked, StructRecordParamTest)
+                self.assertEqual(
+                    unpacked.question,
+                    "The meaning of life, the universe and everything?",
+                )
                 self.assertEqual(unpacked.answer, 42)
+                self.assertEqual(unpacked.needs_clarification, True)
 
     def test_ctype(self):
         extra = None
--- a/comtypes/test/test_midl_safearray_create.py
+++ b/comtypes/test/test_midl_safearray_create.py
@@ -5,21 +5,19 @@ import unittest
 
 import comtypes
 import comtypes.safearray
-from comtypes import CLSCTX_INPROC_SERVER
+from comtypes import CLSCTX_INPROC_SERVER, CLSCTX_LOCAL_SERVER
 from comtypes.client import CreateObject, GetModule
 import comtypes.typeinfo
 
 GetModule("UIAutomationCore.dll")
 from comtypes.gen.UIAutomationClient import CUIAutomation, IUIAutomation
 
-GetModule("scrrun.dll")
-from comtypes.gen.Scripting import Dictionary, IDictionary
-
 ComtypesCppTestSrvLib_GUID = "{07D2AEE5-1DF8-4D2C-953A-554ADFD25F99}"
 
 try:
     GetModule((ComtypesCppTestSrvLib_GUID, 1, 0, 0))
     from comtypes.gen.ComtypesCppTestSrvLib import StructRecordParamTest
+    from comtypes.gen.ComtypesCppTestSrvLib import IDispSafearrayParamTest
 
     IMPORT_FAILED = False
 except (ImportError, OSError):
@@ -43,19 +41,22 @@ class Test_midlSAFEARRAY_create(unittest.TestCase):
                 (unpacked,) = sa.unpack()
                 self.assertIsInstance(unpacked, POINTER(IUIAutomation))
 
+    @unittest.skipIf(IMPORT_FAILED, "This depends on the out of process COM-server.")
     def test_idisp(self):
-        extra = pointer(IDictionary._iid_)
-        idic = CreateObject(Dictionary, interface=IDictionary)
-        idic["foo"] = "bar"
-        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IDictionary))
+        extra = pointer(IDispSafearrayParamTest._iid_)
+        idisp = CreateObject(
+            "Comtypes.DispSafearrayParamTest",
+            clsctx=CLSCTX_LOCAL_SERVER,
+            interface=IDispSafearrayParamTest,
+        )
+        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IDispSafearrayParamTest))
         for ptn, sa in [
-            ("with extra", sa_type.create([idic], extra=extra)),
-            ("without extra", sa_type.create([idic])),
+            ("with extra", sa_type.create([idisp], extra=extra)),
+            ("without extra", sa_type.create([idisp])),
         ]:
             with self.subTest(ptn=ptn):
                 (unpacked,) = sa.unpack()
-                self.assertIsInstance(unpacked, POINTER(IDictionary))
-                self.assertEqual(unpacked["foo"], "bar")
+                self.assertIsInstance(unpacked, POINTER(IDispSafearrayParamTest))
 
     @unittest.skipIf(IMPORT_FAILED, "This depends on the out of process COM-server.")
     def test_record(self):
--- a/comtypes/safearray.py
+++ b/comtypes/safearray.py
@@ -109,7 +109,7 @@ def _make_safearray_type(itemtype):
         _needsfree = False
 
         @classmethod
-        def create(cls, value, extra=None):
+        def create(cls, value, extra=extra):
             """Create a POINTER(SAFEARRAY_...) instance of the correct
             type; value is an object containing the items to store.
 
--- a/source/CppTestSrv/SERVER.IDL
+++ b/source/CppTestSrv/SERVER.IDL
@@ -7,6 +7,8 @@
 
 import "oaidl.idl" ;
 
+// Simple structure used for tests related to IRecordInfo or GetRecordInfoFromGuids functionality.
+// If a new test would require other fields do NOT modify this structure but add a new structure instead.
 typedef [uuid(00FABB0F-5691-41A6-B7C1-11606671F8E5)]
 struct StructRecordParamTest {
 	BSTR question ;


junkmd:
Thank you for your further contribution to this project.

In conclusion, to approve this production codebase, we need to ensure that it does not break the existing codebase that explicitly specifies `extra`, and preventing future regressions where not specifying `extra` would result in an error. In short, we need unit tests for `create`.

> To be quite frank, I don't understand why the extra parameter in the create method is required at all because the method has access to the same closure value. Maybe it's just to keep some flexibility or do I miss something?

I also saw the behavior of `_midlSAFEARRAY` while working on #569, but it is difficult to speculate on the intention behind this implementation.

> As the comment following [line 124 in safearray.py](https://github.com/enthought/comtypes/blob/11d7651e5ed04100ca0388e9617395cfc24092cf/comtypes/safearray.py#L124) says:
> 
> ```
> # For VT_UNKNOWN or VT_DISPATCH, extra must be a pointer to
> # the GUID of the interface.
> #
> # For VT_RECORD, extra must be a pointer to an IRecordInfo
> # describing the record.
> ```

From what I have researched, there are no valid `extra` values other than those mentioned.
However, as you stated,

> while the flexibility to pass a different value for the _extra_ parameter to the _create_ method is maintained.

it is important to maintain backward compatibility.

I understand that there were no direct tests for `create`, and the [`from_param` tests](https://github.com/enthought/comtypes/blob/11d7651e5ed04100ca0388e9617395cfc24092cf/comtypes/test/test_safearray.py#L165) indirectly covered the codebase for `create`.
To confirm the behavior of `create`, unit tests for it are necessary.

The necessary items I thought for unit tests are as follows:
- Definition of a record (a `Structure` with `_recordinfo_` class attribute) and instances to create their arrays
- Definition of a `IUnknown` subclass and instances to create their arrays
- Definition of a `IDispatch` subclass and instances to create their arrays
- Import of `ctypes` types like `c_int` and `c_bool` and instances to create their arrays
- For each case, execute patterns with and without specifying `extra`

For `IUnknown` and `IDispatch`, we can easily obtain subclasses and instances by calling `GetModule`, `CreateObject`, or `oleaut32`-like functions.
For `ctypes` types, simply import and instantiate them.

However, records are not easily prepared.
The only low-cost way currently available in `comtypes` is `ComtypesCppTestSrvLib.StructRecordParamTest`. But, as the name suggests, `ComtypesCppTestSrvLib.StructRecordParamTest` was created for `test_dispifc_records`, and using it for other purposes may complicate dependencies.

If it is not too burdensome for you, I recommend adding a new component to `ComtypesCppTestSrvLib` and using it exclusively for `create`.

Any feedback on this would be welcome.

geppi:
I was afraid that you would ask for unittests. ;-)

I understand that you want to cover non cosmetic changes to the codebase with unittests, although I must say that developing a complete unittest to cover `create` for such a small code change is a little unbalanced. At the end of the day the call to `create` fails with the existing default setting for everything except the basic `ctypes` types. Therefore I would like to limit the effort to the absolute minimum requirement.

We can definitely use the infrastructure provided by the C++ COM test server but I don't think that we require another component. There is no need to interact with a COM server at all in the unittest for `create` of VT_RECORDs. All that is required is a type library that defines some UDTs from which the IRecordInfo can be derived. For this we can use the existing `ComtypesCppTestSrvLib`.

In the cases of _IUnknown_ and _IDispatch_ we might need to use `GetModule `or `CreateObject` but I think can still use existing components.

I'll need to find some time to work on this.

PS.: All would be much easier if we could ask the original author of the package about the intention. :-(

junkmd:
I apologize if my proposal overwhelmed you. I agree that minimal testing should be sufficient.

As the one who made this proposal and as a maintainer, I believe it is my responsibility to specifically explain to contributors what kind of tests should be created.

I wrote these tests by reusing the code I used in interactive mode to verify the actual behavior for static typing. I think using four `itemtype`s is sufficient.

```py
# `test_safearray_create.py`, `test_lp_safearray_create.py` or `test_midl_safearray_create.py`

from ctypes import byref, c_int, pointer, POINTER
import unittest

import comtypes
import comtypes.safearray
from comtypes import CLSCTX_INPROC_SERVER
from comtypes.client import CreateObject, GetModule
import comtypes.typeinfo

GetModule("UIAutomationCore.dll")
from comtypes.gen.UIAutomationClient import CUIAutomation, IUIAutomation

GetModule("scrrun.dll")
from comtypes.gen.Scripting import Dictionary, IDictionary

ComtypesCppTestSrvLib_GUID = "{07D2AEE5-1DF8-4D2C-953A-554ADFD25F99}"

try:
    GetModule((ComtypesCppTestSrvLib_GUID, 1, 0, 0))
    from comtypes.gen.ComtypesCppTestSrvLib import StructRecordSafearrayTest

    IMPORT_FAILED = False
except (ImportError, OSError):
    IMPORT_FAILED = True


class Test_midlSAFEARRAY_create(unittest.TestCase):
    # fixtures...
    def _create_iuia(self) -> IUIAutomation:  # a subclass of `IUnknown`
        return CreateObject(
            CUIAutomation().IPersist_GetClassID(),
            interface=IUIAutomation,
            clsctx=CLSCTX_INPROC_SERVER,
        )

    def _create_sa_iuia_type(self):
        return comtypes.safearray._midlSAFEARRAY(POINTER(IUIAutomation))

    def _create_idic(self) -> IDictionary:  # a subclass of `IDispatch`
        return CreateObject(Dictionary, interface=IDictionary)

    def _create_sa_idic_type(self):
        return comtypes.safearray._midlSAFEARRAY(POINTER(IDictionary))

    def _create_ri(self) -> comtypes.typeinfo.IRecordInfo:
        return comtypes.typeinfo.GetRecordInfoFromTypeInfo(
            *StructRecordSafearrayTest._recordinfo_
        )

    def _create_sa_record_type(self):
        return comtypes.safearray._midlSAFEARRAY(StructRecordSafearrayTest)

    def _create_sa_int_type(self):
        return comtypes.safearray._midlSAFEARRAY(c_int)

    # tests...
    def test_iunk_with_extra(self):
        extra = pointer(IUIAutomation._iid_)
        iuia = self._create_iuia()
        sa = self._create_sa_iuia_type().create([iuia], extra=extra)
        (unpacked,) = sa.unpack()
        self.assertIsInstance(unpacked, POINTER(IUIAutomation))

    def test_iunk_without_extra(self):
        iuia = self._create_iuia()
        sa = self._create_sa_iuia_type().create([iuia])
        (unpacked,) = sa.unpack()
        self.assertIsInstance(unpacked, POINTER(IUIAutomation))

    def test_idisp_with_extra(self):
        extra = pointer(IDictionary._iid_)
        idic = self._create_idic()
        idic["foo"] = "bar"
        sa = self._create_sa_idic_type().create([idic], extra=extra)
        (unpacked,) = sa.unpack()
        self.assertIsInstance(unpacked, POINTER(IDictionary))
        self.assertEqual(unpacked["foo"], "bar")

    def test_idisp_without_extra(self):
        idic = self._create_idic()
        idic["foo"] = "bar"
        sa = self._create_sa_idic_type().create([idic])
        (unpacked,) = sa.unpack()
        self.assertIsInstance(unpacked, POINTER(IDictionary))
        self.assertEqual(unpacked["foo"], "bar")

    @unittest.skipIf(IMPORT_FAILED, "This depends on the out of process COM-server.")
    def test_record_with_extra(self):
        extra = self._create_ri()
        record = StructRecordSafearrayTest()
        sa = self._create_sa_record_type().create([record], extra=extra)
        (unpacked,) = sa.unpack()
        self.assertIsInstance(unpacked, StructRecordSafearrayTest)

    @unittest.skipIf(IMPORT_FAILED, "This depends on the out of process COM-server.")
    def test_record_without_extra(self):
        record = StructRecordSafearrayTest()
        sa = self._create_sa_record_type().create([record])
        (unpacked,) = sa.unpack()
        self.assertIsInstance(unpacked, StructRecordSafearrayTest)

    def test_ctype_with_extra(self):
        extra = None
        cdata = c_int(1)
        sa = self._create_sa_int_type().create([cdata], extra=extra)
        (unpacked,) = sa.unpack()
        self.assertIsInstance(unpacked, int)
        self.assertEqual(unpacked, 1)

    def test_ctype_without_extra(self):
        cdata = c_int(1)
        sa = self._create_sa_int_type().create([cdata])
        (unpacked,) = sa.unpack()
        self.assertIsInstance(unpacked, int)
        self.assertEqual(unpacked, 1)


if __name__ == "__main__":
    unittest.main()


```

I chose `IUIAutomation` as a test double because I am familiar with UIAutomation, but if you know of a more readily available custom COM interface, feel free to use that instead.

> We can definitely use the infrastructure provided by the C++ COM test server, but I don't think that we require another component. There is no need to interact with a COM server at all in the unit test for `create` of VT_RECORDs. All that is required is a type library that defines some UDTs from which the IRecordInfo can be derived. For this, we can use the existing `ComtypesCppTestSrvLib`.

I leave it to you to add record types to `ComtypesCppTestSrvLib`. `StructRecordSafearrayTest` is just a tentative name, so feel free to change it to a better one if you have one in mind.

junkmd:
> PS: All would be much easier if we could ask the original author of the package about the intention. :-(

When maintaining the codebase of this project, I always keep this in mind. The package originator, [Thomas Heller, has stepped down from Python communities](https://blog.python.org/2011/04/thomas-heller-steps-down-as-ctypes.html), and we do not know his current contact information.

In this package, when a `T` type object is passed to a public method, it is often the case that a `_Pointer[T]` is created inside the method and passed to a lower-layer function.

However, I noticed that `create` does not follow this pattern. While writing tests, I found it cumbersome to pass `itemtype` and `extra` as pointers.

I preface this by saying it is just my speculation, but perhaps when `create` was implemented, it was not intended for public use (outside of the `from_param` internals).

However, open-source software often evolves beyond the original creator's intentions, with new use cases discovered by users. If you have a better proposal, please feel free to share it with the community. That will be one step towards the development of this project.

junkmd:
After a night passed, I realized that the test codebase I proposed was a WET despite having many fixtures. When I tried writing it using the subTest approach, the codebase became slimmer. I’ll share it here.

```py
# `test_safearray_create.py`, `test_lp_safearray_create.py` or `test_midl_safearray_create.py`

from ctypes import byref, c_int, pointer, POINTER
import unittest

import comtypes
import comtypes.safearray
from comtypes import CLSCTX_INPROC_SERVER
from comtypes.client import CreateObject, GetModule
import comtypes.typeinfo

GetModule("UIAutomationCore.dll")
from comtypes.gen.UIAutomationClient import CUIAutomation, IUIAutomation

GetModule("scrrun.dll")
from comtypes.gen.Scripting import Dictionary, IDictionary

ComtypesCppTestSrvLib_GUID = "{07D2AEE5-1DF8-4D2C-953A-554ADFD25F99}"

try:
    GetModule((ComtypesCppTestSrvLib_GUID, 1, 0, 0))
    from comtypes.gen.ComtypesCppTestSrvLib import StructRecordSafearrayTest

    IMPORT_FAILED = False
except (ImportError, OSError):
    IMPORT_FAILED = True


class Test_midlSAFEARRAY_create(unittest.TestCase):
    def test_iunk(self):
        extra = pointer(IUIAutomation._iid_)
        iuia = CreateObject(
            CUIAutomation().IPersist_GetClassID(),
            interface=IUIAutomation,
            clsctx=CLSCTX_INPROC_SERVER,
        )
        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IUIAutomation))
        for ptn, sa in [
            ("with extra", sa_type.create([iuia], extra=extra)),
            ("without extra", sa_type.create([iuia])),
        ]:
            with self.subTest(ptn=ptn):
                (unpacked,) = sa.unpack()
                self.assertIsInstance(unpacked, POINTER(IUIAutomation))

    def test_idisp_with_extra(self):
        extra = pointer(IDictionary._iid_)
        idic = CreateObject(Dictionary, interface=IDictionary)
        idic["foo"] = "bar"
        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IDictionary))
        for ptn, sa in [
            ("with extra", sa_type.create([idic], extra=extra)),
            ("without extra", sa_type.create([idic])),
        ]:
            with self.subTest(ptn=ptn):
                (unpacked,) = sa.unpack()
                self.assertIsInstance(unpacked, POINTER(IDictionary))
                self.assertEqual(unpacked["foo"], "bar")

    @unittest.skipIf(IMPORT_FAILED, "This depends on the out of process COM-server.")
    def test_record(self):
        extra = comtypes.typeinfo.GetRecordInfoFromTypeInfo(
            *StructRecordSafearrayTest._recordinfo_
        )
        record = StructRecordSafearrayTest()
        record.answer = 42
        sa_type = comtypes.safearray._midlSAFEARRAY(StructRecordSafearrayTest)
        for ptn, sa in [
            ("with extra", sa_type.create([record], extra=extra)),
            ("without extra", sa_type.create([record])),
        ]:
            with self.subTest(ptn=ptn):
                (unpacked,) = sa.unpack()
                self.assertIsInstance(unpacked, StructRecordSafearrayTest)
                self.assertEqual(unpacked.answer, 42)

    def test_ctype_with_extra(self):
        extra = None
        cdata = c_int(1)
        sa_type = comtypes.safearray._midlSAFEARRAY(c_int)
        for ptn, sa in [
            ("with extra", sa_type.create([cdata], extra=extra)),
            ("without extra", sa_type.create([cdata])),
        ]:
            with self.subTest(ptn=ptn):
                (unpacked,) = sa.unpack()
                self.assertIsInstance(unpacked, int)
                self.assertEqual(unpacked, 1)


if __name__ == "__main__":
    unittest.main()

```

geppi:
Thank you very much for already implementing pretty much the whole unittest.
That is very helpful because I don't really have experience with unittests.

A humble question. Do you really think it is necessary to introduce a new record type for this unittest?

There is no special requirement for the record members, i.e. any record type will do.
Therefore the existing record type `StructRecordParamTest` could be used and the only addition to the codebase would be this single unittest file. If you're concerned about any confusion that could be created by the name, maybe a simple comment in the unittest file would be sufficient that explains the 'reuse' of the existing record type?

junkmd:
Reactions like "Thank you very much" and thumbs up are important interactions in the OSS community to prevent maintainers from burning out.

I also believe that it is an ideal flow to complement each other in the community in areas where we are not good at.

Thank you too.

> There is no special requirement for the record members, i.e. any record type will do.
> Therefore the existing record type `StructRecordParamTest` could be used and the only addition to the codebase would be this single unittest file. If you're concerned about any confusion that could be created by the name, maybe a simple comment in the unittest file would be sufficient that explains the 'reuse' of the existing record type?

I agree with your opinion.
I think that the name `StructRecordParamTest` does not contradict the purpose we are trying to use it for this time.

As you pointed out, I think it would be sufficient if there were comments of this nature before or after the cpp code base shown below.
- This is a simple structures used for tests on functionalities related to `IRecordInfo` or `GetRecordInfoFromGuids`
- If there is a new tests from new perspectives that cannot be covered with the existing fields, we should prepare a new record type rather than changing the field name or type of this record type

https://github.com/enthought/comtypes/blob/11d7651e5ed04100ca0388e9617395cfc24092cf/source/CppTestSrv/SERVER.IDL#L10-L15

geppi:
I'm a little lost as to why this test fails because it passes when I run it in my local environment:
```
python -m unittest .\comtypes\test\test_midl_safearray_create.py
....
Ran 4 tests in 0.021s

OK
```
I'm afraid my current understanding of the inner workings of comtypes is insufficient to make a clue of the error message:
```
AssertionError: <POINTER(IDictionary) ptr=0x1d43dc8 at 4d48440> is not an instance of <class 'comtypes._post_coinit.unknwn.POINTER(IDictionary)'>
```

geppi:
OK, thank you for arranging the Python 3.10 test which shows that the Python version seems not to be the issue.

Well, as I said above, I'm a little clueless what this `comtypes._post_coinit.unknwn.POINTER(IDictionary` type exactly is and why I don't see this error in my local install.

junkmd:
To start with the conclusion, I am not entirely sure why this assertion is causing an error. However, we are not insisting on using `IDictionary` for tests, so I think it would be fine to use an another `IDispatch` instead.

Following your report, I also ran the tests in the GHA environment with `@unittest.skipIf(sys.version_info < (3, x))` added. As a result, even setting x to 12, the `IDictionary` test failed. Therefore, I do not think whether Python is above or below 3.10 is the cause.

I checked whether any `IDispatch` would fail. I imported `IDispSafearrayParamTest` from `ComtypesCppTestSrvLib` (I noticed that the name coincidentally does not contradict the purpose of the test) and ran a test to create an array, which did not fail.

Therefore, I think it would be appropriate to use `IDispSafearrayParamTest` for this test.

```

    def test_idisp(self):
        extra = pointer(IDispSafearrayParamTest._iid_)
        idisp = CreateObject(
            "Comtypes.DispSafearrayParamTest",
            clsctx=CLSCTX_LOCAL_SERVER,
            interface=IDispSafearrayParamTest,
        )
        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IDispSafearrayParamTest))
        for ptn, sa in [
            ("with extra", sa_type.create([idisp], extra=extra)),
            ("without extra", sa_type.create([idisp])),
        ]:
            with self.subTest(ptn=ptn):
                (unpacked,) = sa.unpack()
                self.assertIsInstance(unpacked, POINTER(IDispSafearrayParamTest))

```

junkmd:
Anyway, it's just my assumption, but I think the issue might be due to a broken `IDictionary` in the GHA environment.

If we cannot determine the cause after investigation, I recommend contacting GitHub support.

geppi:
That was a good idea.
All tests pass when using the `IDispSafearrayParamTest` interface instead of `IDictionary` for the dispatch interface test.

junkmd:
I did a bit of research.

I tried using `IFileSystem` defined in `scrrun.dll`, but like `IDictionary`, `self.assertIsInstance(unpacked, POINTER(IFileSystem))` resulted in an error.

```py
# FAILED!
GetModule("scrrun.dll")
from comtypes.gen.Scripting import FileSystemObject, IFileSystem

######
    def test_idisp(self):
        extra = pointer(IFileSystem._iid_)
        idisp = CreateObject(FileSystemObject, interface=IFileSystem)
        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IFileSystem))
        for ptn, sa in [
            ("with extra", sa_type.create([idisp], extra=extra)),
            ("without extra", sa_type.create([idisp])),
        ]:
            with self.subTest(ptn=ptn):
                (unpacked,) = sa.unpack()
                self.assertIsInstance(unpacked, POINTER(IFileSystem))
```

Instead, I tried using `ISpeechVoice` from `sapi.dll` and `assertIsInstance` did not result in an error.

```py
# SUCCEEDED!
GetModule(("{C866CA3A-32F7-11D2-9602-00C04F8EE628}",))  # sapi.dll
from comtypes.gen.SpeechLib import SpVoice, ISpeechVoice

######
    def test_idisp(self):
        extra = pointer(ISpeechVoice._iid_)
        idisp = CreateObject(SpVoice, interface=ISpeechVoice)
        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(ISpeechVoice))
        for ptn, sa in [
            ("with extra", sa_type.create([idisp], extra=extra)),
            ("without extra", sa_type.create([idisp])),
        ]:
            with self.subTest(ptn=ptn):
                (unpacked,) = sa.unpack()
                self.assertIsInstance(unpacked, POINTER(ISpeechVoice))
```

This means that the `scrrun.dll` in the GHA environment might be broken.

In any case, it seems that this issue is originating from the test environment.

geppi:
So from that perspective it's good to have test components available that are under the control of the project.

comtypes/test/test_midl_safearray_create.py
```diff
@@ -0,0 +1,93 @@
...
+                (unpacked,) = sa.unpack()
+                self.assertIsInstance(unpacked, POINTER(IDispSafearrayParamTest))
+
+    @unittest.skipIf(IMPORT_FAILED, "This depends on the out of process COM-server.")
+    def test_record(self):
+        extra = comtypes.typeinfo.GetRecordInfoFromGuids(
+            *StructRecordParamTest._recordinfo_
+        )
+        record = StructRecordParamTest()
+        record.answer = 42
```
junkmd, line 67: I think we should also set some values for `question` and `needs_clarification`, and assert that `unpacked` holds the same values.

junkmd (dismissed):

junkmd:
> So from that perspective it's good to have test components available that are under the control of the project.

That's absolutely true.
Of course, this package intends to interact with 'real' COM type libraries, so tests without test doubles (test components) are important.
However, there are limitations to the COM type libraries that developers and CI automated test environments can prepare.
And there are also environment-caused issues like the ones that occurred this time.

These could be resolved with something like `ComtypesCppTestSrvLib`.

junkmd:
Before I approve this PR, I would like to share the results of my investigation for `IDictionary` with the community.


I wrote a test.
```py
    def test_idic(self):
        extra = pointer(IDictionary._iid_)
        idic = CreateObject(Dictionary, interface=IDictionary)
        idic["foo"] = "bar"
        self.assertIsInstance(idic, IDictionary)
        sa_type = comtypes.safearray._midlSAFEARRAY(POINTER(IDictionary))
        for ptn, sa in [
            ("with extra", sa_type.create([idic], extra=extra)),
            ("without extra", sa_type.create([idic])),
        ]:
            with self.subTest(ptn=ptn):
                (unpacked,) = sa.unpack()
                self.assertEqual(unpacked["foo"], "bar")
                # The `__com_interface__` attribute of a pointer to a COM interface in
                # this package returns the interface class.
                ifc = unpacked.__com_interface__
                print("unpacked.__com_interface__", ifc)
                print("IDictionary", IDictionary)
                self.assertIsInstance(unpacked, ifc)
                self.assertIsInstance(unpacked, POINTER(ifc))
                self.assertIs(ifc, IDictionary)  # FAILED!
                self.assertIsInstance(unpacked, IDictionary)  # FAILED!
                self.assertIsInstance(unpacked, POINTER(IDictionary))  # FAILED!
```

```
======================================================================
FAIL: test_idic (test_midl_safearray_create.Test_midlSAFEARRAY_create.test_idic) (ptn='with extra')
  File "D:\a\comtypes\comtypes\comtypes\test\test_midl_safearray_create.py", line 98, in test_idic
    self.assertIs(ifc, IDictionary)  # FAILED!
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: <class 'comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0.IDictionary'> is not <class 'comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0.IDictionary'>

======================================================================
FAIL: test_idic (test_midl_safearray_create.Test_midlSAFEARRAY_create.test_idic) (ptn='without extra')
  File "D:\a\comtypes\comtypes\comtypes\test\test_midl_safearray_create.py", line 98, in test_idic
    self.assertIs(ifc, IDictionary)  # FAILED!
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: <class 'comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0.IDictionary'> is not <class 'comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0.IDictionary'>

Ran 162 tests in 13.191s

FAILED (failures=2, skipped=29)
unpacked.__com_interface__ <class 'comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0.IDictionary'>
IDictionary <class 'comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0.IDictionary'>
unpacked.__com_interface__ <class 'comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0.IDictionary'>
IDictionary <class 'comtypes.gen._420B2830_E718_11CF_893D_00A0C9054228_0_1_0.IDictionary'>
```

The behavior as a dictionary object is also fine.
The string returned when passed to `repr` is the same.

However, `IDictionary` and `unpacked.__com_interface__` were not related by `is`.

Considering that `ISpeechVoice` which is part of `sapi.dll`, and `IDispSafearrayParamTest` from `ComtypesCppTestSrvLib` do not encounter this problem, I suspect that the `IDictionary` in the GHA environment (possibly the entire `scrrun.dll`) might be broken.

junkmd (approved):
Thank you for your contributing not only for the feature addition but also test code for parts that were previously untested.

@geppi 
If you have the time, I would like you to contact GitHub Support regarding the `IDictionary` problem.

PR #613, opened 2024-09-07
Bump version and update change log for 1.4.7 (and bump `python_requires`)

junkmd:
This is the first version release that supports only Python 3.8 and later.
Following the discussion in https://github.com/enthought/comtypes/issues/216#issuecomment-2154893259, the `z` in the `x.y.z` versioning will be incremented.

After releasing `1.4.7`, Python 3.7 will be removed from the `python-version` in `install-tests` of the CI pipeline.


junkmd:
I am planning to release `comtypes==1.4.7` on September 9th (Japan time).

Issue #512, opened 2024-02-08
Remove remaining `__cmp__` in `comtypes`


junkmd:
In Python3,

> The `cmp()` function should be treated as gone, and the `__cmp__()` special method is no longer supported.
https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons

However, they are still contained in `comtypes`.

https://github.com/enthought/comtypes/blob/afc6488d3386771f69ab4a3e305099e16c34f4d7/comtypes/client/lazybind.py#L129-L132

https://github.com/enthought/comtypes/blob/afc6488d3386771f69ab4a3e305099e16c34f4d7/comtypes/GUID.py#L45-L48

https://github.com/enthought/comtypes/blob/afc6488d3386771f69ab4a3e305099e16c34f4d7/comtypes/__init__.py#L592-L605

Remove these dead code and modernize the codebase.

PR #637, opened 2024-10-11
Deleted `__cmp__` function.

newwingbird:
Related to https://github.com/enthought/comtypes/issues/512.

Since the other PRs seemed to have been neglected, I removed the remaining `__cmp__` methods.

Using ruff on the file.

### DIFF
--- a/comtypes/client/lazybind.py
+++ b/comtypes/client/lazybind.py
@@ -126,11 +126,6 @@ class Dispatch(object):
         "QueryInterface is forwarded to the real com object."
         return self._comobj.QueryInterface(*args)
 
-    def __cmp__(self, other):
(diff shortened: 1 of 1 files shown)


junkmd (approved):
Thanks.

Actually, `_cominterface_meta.__cmp__` still remains.

https://github.com/enthought/comtypes/blob/0256434b2b42e6d67305cf7753a3360d8e3bc345/comtypes/_post_coinit/unknwn.py#L406-L414

The definition of stuffs has moved since I posted #512, but I haven’t updated it.

I would appreciate it if you could address that in another PR.

junkmd:
Since `GUID.__cmp__` was removed in #587, it no longer needs to be addressed.

PR #639, opened 2024-10-11
Removed __cmp__ method

newwingbird:
## Related to under the issue. 
- #512 
- https://github.com/enthought/comtypes/pull/637#pullrequestreview-2362657952

## 
The retained `__cmp__ `method has been deleted
https://github.com/enthought/comtypes/blob/0256434b2b42e6d67305cf7753a3360d8e3bc345/comtypes/_post_coinit/unknwn.py#L406-L414


junkmd (approved):
Thanks.

With this, we no longer need to worry about the linter warnings due to `cmp` not being defined.

Issue #472, opened 2023-01-16
prevent `itf` to be an unbound variable in `_cominterface_meta.__get_baseinterface_methodcount`


junkmd:
Since `itf` can be an unbound variable, the following code will cause several type checkers to warn.

https://github.com/enthought/comtypes/blob/97cbe1c75c3ab514cd643b66520c6c12573c3a28/comtypes/_post_coinit/unknwn.py#L325-L336

But this `try...except... ` clause traps a `KeyError`.

Therefore, it is possible to change to a better code by verifying the existence of the key.

```diff
-        try:
-            result = 0
-            for itf in self.mro()[1:-1]:
+        result = 0
+        for itf in self.mro()[1:-1]:
+            if "_methods_" in itf.__dict__:
                 result += len(itf.__dict__["_methods_"])
-            return result
-        except KeyError as err:
-            (name,) = err.args
-            if name == "_methods_":
-                raise TypeError("baseinterface '%s' has no _methods_" % itf.__name__)
-            raise
+            else:
+                raise TypeError(f"baseinterface '{itf.__name__}' has no _methods_")
+        return result
```

Maybe `vars(itf)` is better than `itf.__dict__`, or maybe there is another way to write it better, but I'll leave it to the contributors for fixing this.

PR #641, opened 2024-10-14
Refactored `itf.__dict__` to `vars(itf)`

newwingbird:
# Proposal

Related to #472

I chose to use `vars` function  because dunder/magic methods serve as the implementation, while invoking functions/methods acts as the API. 

# note 
https://stackoverflow.com/questions/21297203/use-dict-or-vars


junkmd (approved):
These lines used a complex way of catching exceptions to maintain compatibility between Python 2.x and Python 3.x.

Now, the same thing can be written without catching exceptions. Thanks to your efforts, the code has been modernized.

PR #653, opened 2024-11-03
Move the module defining the `GUID` symbol under the `_post_coinit` subpackage.

### LINKED ISSUE #652: Easy to confuse `comtypes.GUID` as a module and as a symbol.
This package has a somewhat confusing namespace where `comtypes.GUID` can refer to either the [module](https://github.com/enthought/comtypes/blob/e62a7e9af0b5ad24aaf9ae8be486d2396e54bac3/comtypes/GUID.py) or the [symbol (Structure)](https://github.com/enthought/comtypes/blob/e62a7e9af0b5ad24aaf9ae8be486d2396e54bac3/comtypes/__init__.py#L38).

Once the `GUID` symbol is imported in `__init__.py`, referencing the `comtypes/GUID.py` module requires metaprogramming with `sys.modules`.

Static type checkers also treat `comtypes.GUID` as a symbol (Structure), not as a module.

This namespace is not particularly readable for developers, even if no problems in runtime.

I think readability would improve if we moved the module under `_post_coinit`, so that `comtypes.GUID` can be recognized only as a symbol.

junkmd:
Closes #652


junkmd:
Tips:
An object equivalent to the `GUID_null` can be easily obtained by instantiating with no arguments passed to the constructor (`GUID()`).

Issue #657, opened 2024-11-04
Refactor from `%` string formatting to f-strings in every module.


junkmd:
We could replace `%` string formatting with f-strings wherever applicable, as was done in #654.

For details on f-string syntax, refer to the [Python official documentation](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals).

- Most of the changes should follow these patterns:
  - Replace `"...%s..." % foo` with `f"...{foo}..."`
  - Replace `"...%r..." % foo` with `f"...{foo!r}..."`
  - Replace `"...%x..." % foo` with `f"...{foo:x}..."`
  - Replace `"...%d..." % foo` with `f"...{foo:d}..."`

- There’s NO need to modify placeholders in logger method calls, as shown below.
  https://github.com/enthought/comtypes/blob/083c19e9fdc0f366df48ad448e897246f346ede0/comtypes/__init__.py#L137-L141

- We welcome keeping changes as small as possible per pull request, as it makes it easier for reviewers to check the changes.

- If your PR addresses multiple rules or files, reviewers might consider it a large change and suggest splitting it into multiple PRs.

newwingbird:
I looked for the use location by git grep -n '%[sdxXorfFeEgGc]' 

I will split modification into several PRs by file or dir. 

<details><summary>result</summary>

```
admin/svndumpfilter2:183:		    self.prop = self.prop + "D %d" % len(key) + "\n" + key + "\n"
admin/svndumpfilter2:185:		    self.prop = self.prop + "K %d" % len(key) + "\n" + key + "\n"
admin/svndumpfilter2:186:		    self.prop = self.prop + "V %d" % len(val) + "\n" + val + "\n"
admin/svndumpfilter2:290:			treecmd = "svnlook tree -r%d %s %s" % \
admin/svndumpfilter2:312:			    props = os.popen("svnlook pl -r%d %s %s" % \
admin/svndumpfilter2:319:				propf = os.popen("svnlook pg -r%d %s %s %s" % \
admin/svndumpfilter2:329:				f = os.popen("svnlook cat -r%d %s %s" % \
comtypes/__init__.py:137:    logger.debug("CoInitializeEx(None, %s)", flags)
comtypes/_comobject.py:79:        "Expected comtypes.COMERROR or WindowsError instance, got %s"
comtypes/_comobject.py:87:        _debug("unimplemented method %s_%s called", interface_name, method_name)
[...]

PR #666, opened 2024-11-16
Refactor string formatting to use f-strings

newwingbird:
Related to #657 

I will splite this modification into some PRs.
This issue's scope is only inside print and string literal.

### DIFF
--- a/comtypes/_memberspec.py
+++ b/comtypes/_memberspec.py
@@ -610,7 +610,7 @@ class bound_named_property(object):
 
     def __iter__(self):
         """Explicitly disallow iteration."""
-        msg = "%r is not iterable" % self.name
+        msg = f"{self.name!r} is not iterable"
         raise TypeError(msg)
 
 
--- a/comtypes/_memberspec.py
+++ b/comtypes/_memberspec.py
@@ -175,15 +175,14 @@ def COMMETHOD(idlflags, restype, methodname, *argspec) -> _ComMemberSpec:
     # join them together(does this make sense?) and replace by None if empty.
     helptext = "".join(t for t in idlflags if isinstance(t, helpstring)) or None
     paramflags, argtypes = _resolve_argspec(argspec)
-    name = (
-        f"_get_{methodname}"
-        if "propget" in idlflags
-        else f"_set_{methodname}"
-        if "propput" in idlflags
-        else f"_setref_{methodname}"
-        if "propputref" in idlflags
-        else methodname
-    )
+    if "propget" in idlflags:
+        name = f"_get_{methodname}"
+    elif "propput" in idlflags:
+        name = f"_set_{methodname}"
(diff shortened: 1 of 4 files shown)


comtypes/automation.py
```diff
@@ -711,17 +711,7 @@ class tagEXCEPINFO(Structure):
...
-                self.wCode,
-                self.bstrSource,
-                self.bstrDescription,
-                self.bstrHelpFile,
-                self.dwHelpContext,
-                self.pfnDeferredFillIn,
-                self.scode,
-            ),
-        )
+        return f"<EXCEPINFO {self.wCode, self.bstrSource, self.bstrDescription, self.bstrHelpFile, self.dwHelpContext, self.pfnDeferredFillIn, self.scode}>"
```
junkmd, line 714: If the line becomes too long, I believe we could make use of a temporary variable.
```suggestion
        info = (self.wCode, self.bstrSource, self.bstrDescription, self.bstrHelpFile, self.dwHelpContext, self.pfnDeferredFillIn, self.scode)
        return f"<EXCEPINFO {info}>"
```
The contents of the tuple will actually be formatted by `ruff`, with appropriate line breaks inserted as needed.

comtypes/_memberspec.py
```diff
@@ -175,14 +175,15 @@ def COMMETHOD(idlflags, restype, methodname, *argspec) -> _ComMemberSpec:
...
-        name = methodname
+    name = (
+        f"_get_{methodname}"
+        if "propget" in idlflags
+        else f"_set_{methodname}"
+        if "propput" in idlflags
+        else f"_setref_{methodname}"
+        if "propputref" in idlflags
+        else methodname
+    )
```
junkmd, line 186: If [conditional expressions (aka "ternary operators")](https://docs.python.org/3/reference/expressions.html#conditional-expressions) becomes long enough that it is broken into multiple lines by the formatter, it is likely more readable to use regular [`if` statements](https://docs.python.org/3/tutorial/controlflow.html#if-statements) instead.

comtypes/_post_coinit/_cominterface_meta_patcher.py
```diff
@@ -17,7 +17,7 @@ def __getattr__(self, name):
             try:
                 fixed_name = self.__map_case__[name.lower()]
             except KeyError:
-                raise AttributeError(name)  # Should we use exception-chaining?
+                raise AttributeError(f"{name}")  # Should we use exception-chaining?
```
junkmd, line 20: Is this necessary?

junkmd (changes requested):

comtypes/automation.py
newwingbird: Good suggestion! 
Using "suggested change" is my first time.

junkmd (approved):
Thanks.

PR #667, opened 2024-11-17
Updating by fstring

newwingbird:
Related to #657 

# Out of scope 

- comtypes/client/__init__.py
- comtypes/server/connectionpoints.py
- comtypes/client/__init__.py
- comtypes/server/connectionpoints.py

## Note 

I investigated by  `git grep -n '%[sdxXorfFeEgGc]' | grep -v 'logger\.debug' | grep -v '_debug' | grep -v '_warning' | grep -v 'logger\.info' | grep -v '_error'`

The following code would be listed with this command.
So Issue's todo list will be changed.
```
            _error(
                "Exception in %s.%s implementation:",
                interface.__name__,
                mthname,
                exc_info=True,
            )
```


comtypes/client/_code_cache.py
```diff
@@ -48,23 +48,23 @@ def _find_gen_dir():
         # check type of executable image to determine a subdirectory
         # where generated modules are placed.
         ftype = getattr(sys, "frozen", None)
-        version_str = "%d%d" % sys.version_info[:2]
+        version_str = f"{sys.version_info[0]:d}{sys.version_info[1]:d}"
         if ftype == None:
             # Python script
-            subdir = r"Python\Python%s\comtypes_cache" % version_str
+            subdir = f"Python\\Python{version_str}\\comtypes_cache"
```
junkmd, line 54: It might be clearer to decompose the [`sys.version_info`](https://docs.python.org/3.13/library/sys.html#sys.version_info) into variables reflecting it as a tuple of *major*, *minor*, *micro*, *releaselevel*, and *serial*.
```diff
-        version_str = f"{sys.version_info[0]:d}{sys.version_info[1]:d}"
+        pymaj, pymin = sys.version_info[:2]
         if ftype == None:
             # Python script
-            subdir = f"Python\\Python{version_str}\\comtypes_cache"
+            subdir = f"Python\\Python{pymaj:d}{pymin:d}\\comtypes_cache"
```

comtypes/server/register.py
```diff
@@ -310,21 +310,21 @@ def _registry_entries(self, cls):
...
                 append(
                     HKCR,
                     "CLSID\\%s\\LocalServer32" % reg_clsid,
                     "",
-                    "%s %s" % (exe, script),
+                    f"{exe} {script}",
                 )
             else:
-                append(HKCR, "CLSID\\%s\\LocalServer32" % reg_clsid, "", "%s" % exe)
+                append(HKCR, "CLSID\\%s\\LocalServer32" % reg_clsid, "", f"{exe}")
```
junkmd, line 327: If you're concerned about the `{}` in f-strings being escaped by a backslash(`\`), there's no need to worry.
As shown below, `f"\\{foo}\\"` would not be escaped. 

```py
>>> from comtypes import GUID
>>> reg_clsid = GUID()
>>> f"CLSID\\{reg_clsid}\\LocalServer32"
'CLSID\\{00000000-0000-0000-0000-000000000000}\\LocalServer32'
>>> "CLSID\\%s\\LocalServer32" % reg_clsid
'CLSID\\{00000000-0000-0000-0000-000000000000}\\LocalServer32'
>>> "CLSID\\%s\\LocalServer32" % reg_clsid == f"CLSID\\{reg_clsid}\\LocalServer32"
True
```

Thus, it seems this could be rewritten as follows. Additionally, there might be other similar parts in the code that can be replaced, so please check them as well. 


```diff
-                append(HKCR, "CLSID\\%s\\LocalServer32" % reg_clsid, "", f"{exe}")
+                append(HKCR, f"CLSID\\{reg_clsid}\\LocalServer32", "", f"{exe}")
```

That said, I've only tested this on Python 3.11, so if it doesn't work with older versions of Python, leaving it as is should be fine.

junkmd (changes requested):

junkmd:
> # Out of scope
> * comtypes/client/**init**.py
> * comtypes/server/connectionpoints.py
> * comtypes/client/**init**.py
> * comtypes/server/connectionpoints.py

Good catch!
I will remove them from the issue kanban.

comtypes/client/_events.py
```diff
@@ -222,12 +222,12 @@ def __getattr__(self, name):
         "Create event handler methods on demand"
         if name.startswith("__") and name.endswith("__"):
             raise AttributeError(name)
-        print("# event found:", name)
+        print(f"# event found: {name}")
```
junkmd, line 225: This part is not related to the issue; rather, it's where spaces are inserted due to the default `sep` of [`print`](https://docs.python.org/3/library/functions.html#print).

Please revert this line to its original.

junkmd (changes requested):

comtypes/server/register.py
```diff
@@ -310,18 +310,18 @@ def _registry_entries(self, cls):
...
+                    script = f'"{script}"'
                 append(
                     HKCR,
                     "CLSID\\%s\\LocalServer32" % reg_clsid,
                     "",
-                    "%s %s" % (exe, script),
+                    f"{exe} {script}",
                 )
             else:
                 append(HKCR, "CLSID\\%s\\LocalServer32" % reg_clsid, "", "%s" % exe)
```
junkmd, line 327:
```py
                append(HKCR, f"CLSID\\{reg_clsid}\\LocalServer32", "", f"{exe}")
```

comtypes/client/_code_cache.py
```diff
@@ -48,23 +48,23 @@ def _find_gen_dir():
         # check type of executable image to determine a subdirectory
         # where generated modules are placed.
         ftype = getattr(sys, "frozen", None)
-        version_str = "%d%d" % sys.version_info[:2]
+        pymaj, pymin = sys.version_info[:2]
         if ftype == None:
             # Python script
-            subdir = r"Python\Python%s\comtypes_cache" % version_str
+            subdir = rf"Python\\Python{pymaj:d}{pymin:d}\\comtypes_cache"
```
junkmd, line 54: `/` might be excessive.

comtypes/server/register.py
```diff
@@ -310,18 +310,18 @@ def _registry_entries(self, cls):
...
                 if not __debug__:
-                    exe = "%s -O" % exe
+                    exe = f"{exe} -O"
                 script = os.path.abspath(sys.modules[cls.__module__].__file__)
                 if " " in script:
-                    script = '"%s"' % script
+                    script = f'"{script}"'
                 append(
                     HKCR,
                     "CLSID\\%s\\LocalServer32" % reg_clsid,
```
junkmd, line 322:
```py
                    f"CLSID\\{reg_clsid}\\LocalServer32",
```

junkmd (changes requested):
`append(…` could likely be replaced with f-strings for other lines as well.

junkmd:
To avoid further confusion while working on this issue, I’ve added examples to the issue kanban.

> * Replace `"...\\%s..." % foo` with `f"...\\{foo}..."`
> * Replace `r"...\%s..." % foo` with `rf"...\{foo}..."`


https://github.com/enthought/comtypes/issues/657#issuecomment-2481672206

comtypes/server/register.py
```diff
@@ -310,21 +310,16 @@ def _registry_entries(self, cls):
...
                 if " " in script:
-                    script = '"%s"' % script
-                append(
-                    HKCR,
-                    "CLSID\\%s\\LocalServer32" % reg_clsid,
-                    "",
-                    "%s %s" % (exe, script),
-                )
+                    script = f'"{script}"'
+                append(HKCR, rf"CLSID\{reg_clsid}\LocalServer32", "", f"{exe}")
```
junkmd, line 320: Originally, it was as below, so `f"{exe}"` needs to be changed to `f"{exe} {script}"`.

https://github.com/enthought/comtypes/blob/1ea74f5a68c4f1e0ee511ba3591e2204688ec071/comtypes/server/register.py#L318-L327
```diff
-               append(HKCR, rf"CLSID\{reg_clsid}\LocalServer32", "", f"{exe}")
+               append(HKCR, rf"CLSID\{reg_clsid}\LocalServer32", "", f"{exe} {script}")
```

junkmd (changes requested):

junkmd (changes requested):

junkmd (approved):
Thanks for your efforts.

I’ve reviewed several PRs related to #657 and have come to the conclusion that the optimal amount of changes for this type of replacement is about 30 lines per PR.

Of course, there may be cases related to #657 where changes exceeding 30 lines are necessary for a single file. In such cases, please create a PR for that file only.

PR #677, opened 2024-12-02
Fix broken unregistration of 64bit type libraries

forderud:
The `UnRegisterTypeLib` function currently assume 32bit type libraries (`SYS_WIN32`) if called _without_ explicit `syskind` argument. This is the case when UnRegisterTypeLib is called implicitly from `comtypes.server.register.UseCommandLine`, since the `_reg_typelib_` field only contains the first 3 `libID` , `wVerMajor` and `wVerMinor` arguments.

Propose to fix the problem by updating the hardcoding to 64bit Windows with `SYS_WIN64`. This will fix the problem in 64bit at the expense of breaking unregistration of 32bit type libraries.

## Alternative approach
It would also be possible to instead extend the [GetLibAttr()](https://learn.microsoft.com/en-us/windows/win32/api/oaidl/nf-oaidl-itypelib-getlibattr) parsing to also capture the `lcid` and `syskind` parameters from the [TLIBATTR](https://learn.microsoft.com/en-us/windows/win32/api/oaidl/ns-oaidl-tlibattr) struct, so that they can be forwarded to `UnRegisterTypeLib` without having to rely on default values.
I assume that such a change will need to be implemented in [tlbparser.py](https://github.com/enthought/comtypes/blob/main/comtypes/tools/tlbparser.py). However, I don't really know the codebase myself, so I don't feel comfortable with coming up with a concrete proposal.

## Example of impact
The https://github.com/forderud/ComSamples/blob/main/MyServerPy/MyServerPy.py project is experiencing this problem and have implemented a work-around for it.

### DIFF
--- a/comtypes/typeinfo.py
+++ b/comtypes/typeinfo.py
@@ -632,7 +632,7 @@ def LoadTypeLib(szFile: str) -> ITypeLib:
 
 
 def UnRegisterTypeLib(
-    libID: str, wVerMajor: int, wVerMinor: int, lcid: int = 0, syskind: int = SYS_WIN32
+    libID: str, wVerMajor: int, wVerMinor: int, lcid: int = 0, syskind: int = SYS_WIN64
 ) -> int:
     """Unregister a registered type library"""
     return _oleaut32.UnRegisterTypeLib(


junkmd:
Thank you for your contribution.

I think the changes proposed in this PR should be avoided because it breaks backward compatibility, as you pointed out.  
Additionally, modifying `tlbparser` might affect the  `codegenerator`, which depends on it.

If we were to obtain `lcid` and `syskind`, my idea is to add optional entries to `CoClassImpl` and then introduce a conditional branches in `comtypes/server/register.py` to retrieve those values, as shown below.

```diff
 class MyServerImpl(MyInterfaces.MyServer):
      # registry entries
+    _reg_lcid_ = 0
+    _reg_syskind_ = SYS_WIN64
     _reg_threading_ = "Both"
     _reg_desc_ = "Python-based COM server for testing"
     _reg_clsctx_ = comtypes.CLSCTX_LOCAL_SERVER # run in separate process
```

```diff
@@ -42,7 +42,12 @@ import winreg
 import logging
 
 import comtypes
-from comtypes.typeinfo import LoadTypeLibEx, UnRegisterTypeLib, REGKIND_REGISTER
+from comtypes.typeinfo import (
+    LoadTypeLibEx,
+    UnRegisterTypeLib,
+    REGKIND_REGISTER,
+    SYS_WIN32,
+)
 from comtypes.hresult import *
 from comtypes.server import w_getopt
 import comtypes.server.inprocserver
@@ -202,10 +207,12 @@ class Registrar(object):
                 if get_winerror(detail) != 2:
                     raise
         tlib = getattr(cls, "_reg_typelib_", None)
+        lcid = getattr(cls, "_reg_lcid_", 0)
+        syskind = getattr(cls, "_reg_syskind_", SYS_WIN32)
         if tlib is not None:
             try:
                 _debug("UnRegisterTypeLib(%s, %s, %s)", *tlib)
-                UnRegisterTypeLib(*tlib)
+                UnRegisterTypeLib(*tlib, lcid, syskind)
             except WindowsError as detail:
                 if not get_winerror(detail) in (
                     TYPE_E_REGISTRYACCESS,
@@ -254,6 +261,8 @@ class Registrar(object):
         _reg_progid_ - a string naming the progid, typically 'MyServer.MyObject.1'
         _reg_novers_progid_ - version independend progid, typically 'MyServer.MyObject'
         _reg_typelib_ - an tuple (libid, majorversion, minorversion) specifying a typelib.
+        _reg_lcid_ - ...
+        _reg_syskind - ...
         _reg_threading_ - a string specifying the threading model
 
         Note that the first part of the progid string is typically the
```

However, this area has long been without documentation updates, and its tests are broken.  
Whichever approach we choose, I believe we should carefully discuss, verify, and adopt the most reasonable one agreed upon by the community.

junkmd:
@bennyrowland  
I would like to hear your thoughts on whether this discussion is reasonable, especially considering your experience implementing local servers using this package.

Given the current dominance of Win64, it might be worth considering whether server-side implementation requires more significant changes—not just quick fixes like this one—but at a level that justifies a version update to `1.5.0`.

forderud:
One alternative thought @junkmd: Could it make sense to instead update `UnRegisterTypeLib` to default to `SYS_WIN32` or `SYS_WIN64` depending on the "bitness" of the Python interpreter?

A 32bit Python COM servers will anyhow not be compatible with a 64bit Python installation and vice versa. We'll then avoid breaking existing 32bit COM servers running from a 32bit Python interpreter.

forderud:
I've now submitted an updated proposal #678 that passes `SYS_WIN64` or `SYS_WIN32` depending of the "bitness" of the Python interpreter. Therefore, please feel free to close this PR.

forderud:
> ```diff
>  class MyServerImpl(MyInterfaces.MyServer):
>       # registry entries
> +    _reg_lcid_ = 0
> +    _reg_syskind_ = SYS_WIN64
>      _reg_threading_ = "Both"
>      _reg_desc_ = "Python-based COM server for testing"
>      _reg_clsctx_ = comtypes.CLSCTX_LOCAL_SERVER # run in separate process
> ```

@junkmd The "lcid" and "syskind" properties are really tied to the type library (TLB) and not to the COM class. Having to hardcode the "bitness" with SYS_WIN64 in the python sources is also suboptimal, since it makes it klunky to support both 32 & 64bit from a single implementation. I therefore think it would be more natural to instead extend `_reg_typelib_`  with additional fields for this that could potentially be optional.

bennyrowland:
@forderud the new fix you have provided is exactly that which I was planning to propose, thanks for implementing.

junkmd:
@bennyrowland 
Thank you for joining this discussion and your insights.

@forderud 
> I've now submitted an updated proposal #678 that passes `SYS_WIN64` or `SYS_WIN32` depending of the "bitness" of the Python interpreter. Therefore, please feel free to close this PR.

Alright, let’s continue this topic in #678.

PR #692, opened 2024-12-08
Move win32com stuff from `test_dispinterface` to `test_comserver_pywin32`.

junkmd:


junkmd:
It didn’t work, so I will reconsider the approach.

PR #756, opened 2025-01-25
[typeinfo.py] Use WinDLL/OleDLL instead of windll/oledll

moi15moi:
Note that I removed the check to see if oleaut32 has the "LoadTypeLibEx" function. From what I can see, the function is available since Windows 2000, so we don't need this check.

### DIFF
--- a/comtypes/typeinfo.py
+++ b/comtypes/typeinfo.py
@@ -6,7 +6,17 @@
 import ctypes
 import sys
 import weakref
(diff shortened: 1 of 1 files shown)


junkmd:
> Note that I removed the check to see if oleaut32 has the "LoadTypeLibEx" function. From what I can see, the function is available since Windows 2000, so we don't need this check.

Understood.
I will include this in the changelog as well.

junkmd (approved):
Thanks.

PR #763, opened 2025-01-25
[tools\tlbparser.py] Import QueryPathOfRegTypeLib from typeinfo + doc adjustments

moi15moi:
See #735

I needed to use a try except because, now, _QueryPathOfRegTypeLib return a HRESULT and not a c_int

### DIFF
--- a/comtypes/tools/tlbparser.py
+++ b/comtypes/tools/tlbparser.py
@@ -1,6 +1,6 @@
 import os
 import sys
-from ctypes import alignment, byref, c_void_p, sizeof, windll
+from ctypes import alignment, c_void_p, sizeof, windll
 from typing import Any, Dict, List, Optional, Tuple
 
 from comtypes import BSTR, COMError, automation, typeinfo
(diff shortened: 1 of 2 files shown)


comtypes/tools/tlbparser.py
```diff
@@ -740,20 +741,23 @@ def get_tlib_filename(tlib: typeinfo.ITypeLib) -> Optional[str]:
...
-    return None
+    try:
+        _QueryPathOfRegTypeLib(
+            byref(la.guid), la.wMajorVerNum, la.wMinorVerNum, 0, byref(name)
+        )
+    except OSError:
+        return None
+
+    full_filename = name.value.split("\0")[0]
+    if not os.path.isabs(full_filename):
```
junkmd, line 752: If making such changes, should we now be using higher-level public functions?  
`QueryPathOfRegTypeLib` eliminates the need to prepare a `BSTR` and the handling of the null terminator.

```diff
 def get_tlib_filename(tlib: typeinfo.ITypeLib) -> Optional[str]:
     # seems if the typelib is not registered, there's no way to
     # determine the filename.
     la = tlib.GetLibAttr()
-    name = BSTR()
     try:
-        _QueryPathOfRegTypeLib(
-            byref(la.guid), la.wMajorVerNum, la.wMinorVerNum, 0, byref(name)
+        full_filename = typeinfo.QueryPathOfRegTypeLib(
+            str(la.guid), la.wMajorVerNum, la.wMinorVerNum
         )
     except OSError:
         return None

-    full_filename = name.value.split("\0")[0]
     if not os.path.isabs(full_filename):
         # workaround Windows 7 bug in QueryPathOfRegTypeLib returning relative path
         try:
```

Since these changes exceed the scope of #735, I believe the PR title could be updated to better focus on `QueryPathOfRegTypeLib`.

P.S. Import part:
```diff
 import os
 import sys
-from ctypes import alignment, byref, c_void_p, sizeof, windll
+from ctypes import alignment, c_void_p, sizeof, windll
 from typing import Any, Dict, List, Optional, Tuple

-from comtypes import BSTR, COMError, automation, typeinfo
+from comtypes import COMError, automation, typeinfo
 from comtypes.client._code_cache import _get_module_filename
 from comtypes.tools import typedesc
-from comtypes.typeinfo import _QueryPathOfRegTypeLib
```

junkmd (changes requested):
This goes beyond simply replacing `windll` with `WinDLL` and likely exceeds the scope of #735.
However, the concept behind this change is excellent as a refactoring effort, so let’s keep this PR open and continue the discussion about `QueryPathOfRegTypeLib`.

(This is one of the benefits of [Small CLs](https://google.github.io/eng-practices/review/developer/small-cls.html). If there were other types of changes or modifications to other modules, this PR would have taken much longer to reach the point of merging.)

comtypes/tools/tlbparser.py
moi15moi: You right.
Now, I directly call ``QueryPathOfRegTypeLib`` from ``typeinfo``.
I also added some documentation, so we can clearly understand that ``QueryPathOfRegTypeLib`` can raise a OSError.

junkmd (approved):
Thanks!

PR #764, opened 2025-01-25
[test\test_basic.py] Use OleDLL instead of windll

moi15moi:
See #735

IUnknown to ICreateTypeLib2 because CreateTypeLib2 doesn't take IUnknown as a parameter.

### DIFF
--- a/comtypes/test/test_basic.py
+++ b/comtypes/test/test_basic.py
@@ -1,7 +1,8 @@
 ##import ut
 import unittest as ut
-from ctypes import windll, POINTER, byref, HRESULT
+from ctypes import POINTER, byref, HRESULT
 from comtypes import IUnknown, STDMETHOD, GUID
+from comtypes.typeinfo import _CreateTypeLib2, ICreateTypeLib2
 
 # XXX leaks references!
 
@@ -20,8 +21,10 @@ class BasicTest(ut.TestCase):
         POINTER(IUnknown)()
 
     def test_refcounts(self):
-        p = POINTER(IUnknown)()
-        windll.oleaut32.CreateTypeLib2(1, "blabla", byref(p))
+        # Since all COM interfaces derive from IUnknown and have the same reference counting behavior, any interface
+        # — whether ICreateTypeLib2 or otherwise — could be used for this test.
+        p = POINTER(ICreateTypeLib2)()
+        _CreateTypeLib2(1, "blabla", byref(p))
         # initial refcount is 2
         for i in range(2, 10):
             self.assertEqual(p.AddRef(), i)
@@ -29,8 +32,10 @@ class BasicTest(ut.TestCase):
             self.assertEqual(p.Release(), i)
 
     def test_qi(self):
-        p = POINTER(IUnknown)()
-        windll.oleaut32.CreateTypeLib2(1, "blabla", byref(p))
+        # Since all COM interfaces derive from IUnknown and have the same QueryInterface behavior, any interface
+        # — whether ICreateTypeLib2 or otherwise — could be used for this test.
+        p = POINTER(ICreateTypeLib2)()
+        _CreateTypeLib2(1, "blabla", byref(p))
         self.assertEqual(p.AddRef(), 2)
         self.assertEqual(p.Release(), 1)
 


junkmd (commented):
Here’s my perspective on this test:

Given the objectives of `test_refcounts` and `test_qi`, any test double that serves as a pointer to `IUnknown` should suffice. However, I think it's better to avoid approaches that rely on specific custom interface definitions. 
Since `test_identity` calls `CreateTypeLib` (that indirectly calls `_CreateTypeLib2`), I don’t see much value in testing `_CreateTypeLib2` bound to `ICreateTypeLib2` specifically in this context. 

https://github.com/enthought/comtypes/blob/a9f0558e4d5ce1476f6a330127c910f805e43274/comtypes/test/test_basic.py#L101-L115


It might be worth considering defining a single-responsibility function to create this test double, duplicating `_CreateTypeLib2` from `typeinfo.py` and make the “`_CreateTypeLib2` that accepts `POINTER(IUnknown)`”.

What are your thoughts on this?

moi15moi:
I prefer to adhere strictly to the actual parameters of [[CreateTypeLib2](https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-createtypelib2)](https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-createtypelib2) to maintain accuracy and consistency.

Since `IUnknown` behaves like an abstract class, there is no direct way to test `refcounts` and `QueryInterface` and I think it's totally normal. However, I believe it would be beneficial to include a comment indicating that the purpose is to test the behavior of `IUnknown` even if we use ``ICreateTypeLib2``.

What do you think?

junkmd:
> However, I believe it would be beneficial to include a comment indicating that the purpose is to test the behavior of `IUnknown` even if we use `ICreateTypeLib2`.

Indeed, comments can be a powerful tool in situations like this.  

Considering a scenario where future maintainers might need to modify these tests, it’s important to document our thought process here.  

Please make a commit to add comments.

moi15moi:
> Please make a commit to add comments.

Done, let me know what you think.

comtypes/test/test_basic.py
```diff
@@ -20,17 +21,21 @@ def test_release(self):
         POINTER(IUnknown)()
 
     def test_refcounts(self):
-        p = POINTER(IUnknown)()
-        windll.oleaut32.CreateTypeLib2(1, "blabla", byref(p))
+        # Since IUnknown behaves like an abstract class, the reference counting behavior cannot be tested directly.
```
junkmd, line 24: As the previous code demonstrates, `QueryInterface` and refcount tests can be performed with this package’s `IUnknown`.

To better reflect the accurate situation, I think a comments like “COM interfaces can use `AddRef`, `Release`, and `QueryInterface` inherited from `IUnknown`” would be more appropriate.

junkmd:
```diff
-        # Since IUnknown behaves like an abstract class, the reference counting behavior cannot be tested directly.
-        # Although we use ICreateTypeLib2 here, the aim is to validate the reference counting behavior of IUnknown.
+        # Since all COM interfaces derive from IUnknown and have the same reference counting behavior, any interface
+        # — whether ICreateTypeLib2 or otherwise — could be used for this test.
```

Would this be sufficient?
A comment like this would help future maintainers understand that this test is NOT strictly tied to `ICreateTypeLib2`.

junkmd (changes requested):

comtypes/test/test_basic.py
```diff
@@ -20,17 +21,21 @@ def test_release(self):
...
         for i in range(2, 10):
             self.assertEqual(p.AddRef(), i)
         for i in range(8, 0, -1):
             self.assertEqual(p.Release(), i)
 
     def test_qi(self):
-        p = POINTER(IUnknown)()
-        windll.oleaut32.CreateTypeLib2(1, "blabla", byref(p))
+        # Since IUnknown behaves like an abstract class, the reference counting behavior cannot be tested directly.
+        # Although we use ICreateTypeLib2 here, the aim is to validate the reference counting behavior of IUnknown.
```
junkmd, line 36:
```diff
-        # Since IUnknown behaves like an abstract class, the reference counting behavior cannot be tested directly.
-        # Although we use ICreateTypeLib2 here, the aim is to validate the reference counting behavior of IUnknown.
+        # Since all COM interfaces derive from IUnknown and have the same interface querying behavior, any interface
+        # — whether ICreateTypeLib2 or otherwise — could be used for this test.
```

junkmd (changes requested):

comtypes/test/test_basic.py
moi15moi: I applied your suggestion

comtypes/test/test_basic.py
moi15moi: I applied your suggestion

junkmd (approved):
Thanks!

PR #765, opened 2025-01-25
[test\test_outparam.py] Use WinDLL/OleDLL instead of windll/oledll

moi15moi:
See #735

### DIFF
--- a/comtypes/test/test_outparam.py
+++ b/comtypes/test/test_outparam.py
@@ -1,23 +1,26 @@
 import unittest
 from ctypes import (
+    HRESULT,
     POINTER,
+    OleDLL,
+    WinDLL,
     byref,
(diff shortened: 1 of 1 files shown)


junkmd (approved):
`CoTaskMemFree` is not solely for the `GUID` structure, so it might be better to define it in `__init__.py` or a new module rather than in `comtypes/GUID.py`.
Considering that `CoTaskMemAlloc`, `IMalloc`, `CoGetMalloc` (and potentially `CoTaskMemRealloc`) are valuable as COM utilities, there could be a dedicated module for these utilities.

That said, implementing such changes would exceed the scope of #735, so we should avoid addressing it here.
This is not something that needs to be resolved immediately, but rather a future consideration.

PR #766, opened 2025-01-25
[test\test_urlhistory.py] Use WinDLL instead of windll

moi15moi:
See #735


junkmd (approved):
What I want to say about `CoTaskMemFree` is the same as https://github.com/enthought/comtypes/pull/765#pullrequestreview-2574108723.

junkmd:
Thanks.

PR #767, opened 2025-01-25
[test\test_comobject.py] Use OleDLL instead of oledll

moi15moi:
See #735


junkmd (approved):
Thanks!

Regarding importing `_CoCreateInstance` from `_post_coinit.misc`: This is currently the only way to access this private function, and since this is part of the test codebase, I have determined that the level of redundancy required for production codebases is not necessary at this time.

PR #759, opened 2025-01-25
[server\localserver.py] Use OleDLL instead of oledll

moi15moi:
See #735

PR #769, opened 2025-01-25
[server\__init__.py] Use OleDLL instead of oledll

moi15moi:
See #735

### DIFF
--- a/comtypes/server/__init__.py
+++ b/comtypes/server/__init__.py
@@ -1,5 +1,6 @@
 import ctypes
-from ctypes import HRESULT, POINTER, byref
+from ctypes import HRESULT, POINTER, OleDLL, byref, c_void_p
+from ctypes.wintypes import DWORD, LPVOID
 from typing import TYPE_CHECKING, Any, Optional, Type
 
 import comtypes
@@ -7,6 +8,7 @@ import comtypes.client
 import comtypes.client.dynamic
 from comtypes import GUID, STDMETHOD, IUnknown
 from comtypes.automation import IDispatch
+from comtypes.GUID import REFCLSID
 
 if TYPE_CHECKING:
     from ctypes import _Pointer
@@ -67,7 +69,23 @@ class IClassFactory(IUnknown):
 ACTIVEOBJECT_STRONG = 0x0
 ACTIVEOBJECT_WEAK = 0x1
 
-oleaut32 = ctypes.oledll.oleaut32
+_oleaut32 = OleDLL("oleaut32")
+
+_RegisterActiveObject = _oleaut32.RegisterActiveObject
+_RegisterActiveObject.argtypes = [
+    c_void_p,
+    REFCLSID,
+    DWORD,
+    POINTER(DWORD),
+]
+_RegisterActiveObject.restype = HRESULT
+
+_RevokeActiveObject = _oleaut32.RevokeActiveObject
+_RevokeActiveObject.argtypes = [
+    DWORD,
+    LPVOID,
+]
+_RevokeActiveObject.restype = HRESULT
 
 
 def RegisterActiveObject(comobj: comtypes.COMObject, weak: bool = True) -> int:
@@ -78,9 +96,9 @@ def RegisterActiveObject(comobj: comtypes.COMObject, weak: bool = True) -> int:
     else:
         flags = ACTIVEOBJECT_STRONG
     handle = ctypes.c_ulong()
-    oleaut32.RegisterActiveObject(punk, byref(clsid), flags, byref(handle))
+    _RegisterActiveObject(punk, byref(clsid), flags, byref(handle))
     return handle.value
 
 
 def RevokeActiveObject(handle: int) -> None:
-    oleaut32.RevokeActiveObject(handle, None)
+    _RevokeActiveObject(handle, None)


comtypes/server/__init__.py
```diff
@@ -67,7 +69,23 @@ def LockServer(self, fLock: int) -> hints.Hresult: ...
...
-oleaut32 = ctypes.oledll.oleaut32
+_oleaut32 = OleDLL("oleaut32")
+
+_RegisterActiveObject = _oleaut32.RegisterActiveObject
+_RegisterActiveObject.argtypes = [
+    POINTER(IUnknown),
+    REFCLSID,
+    DWORD,
+    POINTER(DWORD),
+]
```
junkmd, line 80: I believe this has the same issue as https://github.com/enthought/comtypes/pull/759#discussion_r1929525378.
The reason the CI is passing is that there are no tests for this function.

I recommend changing `POINTER(IUnknown)` to `c_void_p`.

junkmd (changes requested):

comtypes/server/__init__.py
moi15moi: > I recommend changing `POINTER(IUnknown)` to `c_void_p`.

Before doing that, I think it would be worth it to create a test that would cover the ``RegisterActiveObject`` and ``RevokeActiveObject`` function, but I have no knownledge about comtype server.

So, I'll keep this PR open so that if tests for these functions are added in the future, we can continue the investigation

junkmd: Agreed.

Without untangling the server-side implementation, progress on python/cpython#127369 cannot be made.

There is documentation on COM/OLE published under the MIT license.
https://github.com/kraigb/InsideOLE/

If you can allocate resources to this issue and learn something from reading that documentation, please feel free to share it.

junkmd: This is what I know about these function and my observations.

These are _simple wrapper functions around the `GetRunningObjectTable` API and the `IRunningObjectTable` interface_ [^1].

[`RegisterActiveObject`](https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-registeractiveobject) corresponds [`IRunningObjectTable::Register`](https://learn.microsoft.com/en-us/windows/win32/api/objidl/nf-objidl-irunningobjecttable-register).
[`RevokeActiveObject`](https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-revokeactiveobject) corresponds [`IRunningObjectTable::Revoke method`](https://learn.microsoft.com/en-us/windows/win32/api/objidl/nf-objidl-irunningobjecttable-revoke).

The tests for `IRunningObjectTable` is existing in this project.
https://github.com/enthought/comtypes/blob/89454467c041466dc8077c15d82248cb7578778c/comtypes/test/test_monikers.py#L57-L70

From the changelog, [they were added without any tests](https://github.com/enthought/comtypes/commit/ea3c333c01674b28557513cd3e47ebe3e3f1261c), and I can’t find any use cases for these functions in the server.

https://github.com/enthought/comtypes/blob/89454467c041466dc8077c15d82248cb7578778c/CHANGES.txt#L624-L627

It’s puzzling that these functions are not defined near `GetActiveObject` and instead reside on the server side.
Additionally, it’s strange that `RegisterActiveObject`, which is determining which instance is the active object of a class, are not wrapped to accept `punk`.

Perhaps these functions might be marked as deprecated and adding alternative functions somewhere else in this project.

What do you think?

[^1]: [InsideOLE.pdf](https://github.com/kraigb/InsideOLE/blob/main/InsideOLE.pdf), p489.

junkmd: I’ve figured out the use case for `RegisterActiveObject`.

When we pass a `SomeComObject` instance (let’s say it's assigned to the variable `c`) to this function (`RegisterActiveObject(c)`), the instance `o` obtained via `o = c.QueryInterface(ISomeComInterface)` becomes the active object of that class.
We can then retrieve this instance using `p = GetActiveObject(SomeComObject)`, and the condition `o == p` holds true.

This behavior is indeed meaningful as a utility function.

We’ve written a test for this and will submit a PR.

(P.S. That said, I think having a version of `RegisterActiveObject` that accepts `punk` would also be convenient. In the future, we might see the addition of a `RegisterActiveObject` on the `client` side that accepts arguments differing from those on the `server` side.)

junkmd: I opened https://github.com/enthought/comtypes/pull/776.

junkmd: #776 has been merged.

Please rebase this PR’s branch and check the CI pipeline test results.

moi15moi: Is there any doc to build ``source\CppTestSrv``?
I tried with Visual Studio, but when I run the command ``server.exe /RegServer``, [this assert](https://github.com/enthought/comtypes/blob/d3057130edf42bb5312405f63559b2224a26b45c/source/CppTestSrv/REGISTRY.CPP#L145) always fails.

moi15moi: Like you said, I changed ``POINTER(IUnknown)`` to ``c_void_p`` for ``_RegisterActiveObject``.

To know why, see this pipeline where I used ``POINTER(IUnknown)``: https://github.com/enthought/comtypes/actions/runs/13022699670/job/36326427772

junkmd: > Is there any doc to build `source\CppTestSrv`?

There is no documentation on building the test doubles, but I believe it might work if you replicate the CI pipeline workflow.

https://github.com/enthought/comtypes/blob/d3057130edf42bb5312405f63559b2224a26b45c/.github/workflows/autotest.yml#L23-L29

Comparing the Visual Studio setup provided by [`ilammy/msvc-dev-cmd`](https://github.com/ilammy/msvc-dev-cmd) with your environment might reveal some insights.

junkmd (approved):
Thanks!

PR #773, opened 2025-01-26
Replace deprecated install command

moi15moi:
See: https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-commands-should-be-used-instead


junkmd (approved):
Thanks!

Issue #772, opened 2025-01-26
Use coverage in the CI


moi15moi:
I think it would be beneficial to use [coverage](https://coverage.readthedocs.io/) to track which part of comtypes is covered by tests.

The report created by coverage could be published on codecov.io.
In our CI, we could create the report and use [codecov/codecov-action](https://github.com/codecov/codecov-action) to publish it.

Here is a example of a project on codecov.io: https://app.codecov.io/github/moi15moi/FontCollector
As you can see, you can clearly see which part is covered and which part isn't which helps a lot to create some tests.

Note that I am note sure if it will properly detect the code covered in the [server](https://github.com/enthought/comtypes/tree/9e9a8aad071d9058444f3c0a12c6c5237c321ffe/comtypes/server), but I still think it would be a plus.

junkmd:
That's a great idea!

First, I’d appreciate it if you could test and verify what the appropriate codecov.io configuration for this project would be within your forked repository.

If applying codecov.io requires repository owner permissions, please let me know. I’m considering reaching out to Enthought, the host of this repository, to either allocate the necessary permissions or make the required configuration changes.

moi15moi:
Just so you know, this issue is currently blocked by this bug: https://github.com/nedbat/coveragepy/issues/1920

PR #775, opened 2025-01-27
Add support for coverage

moi15moi:
I opened this PR as a draft because I still haven't implement everthing said in #772.

Also, now, to run the tests, I use ``coverage run -m unittest discover -v -s comtypes\test -t comtypes\test`` instead of ``python -m unittest discover -v -s ./comtypes/test -t comtypes\test``, but ``test_eval`` consistently fails with a large memory leaks:
```py
FAIL: test_eval (test_comserver.TestInproc)
  File "D:\a\comtypes\comtypes\comtypes\test\test_comserver.py", line 99, in test_eval
    self._find_memleak(func)
  File "D:\a\comtypes\comtypes\comtypes\test\test_comserver.py", line 42, in _find_memleak
    self.assertFalse(bytes, f"Leaks {bytes} bytes")  # type: ignore
AssertionError: 5591040 is not false : Leaks 5591040 bytes
```

Have you an idea why it happen?
We might be hitting a element from [Things that don’t work](https://coverage.readthedocs.io/en/7.6.10/trouble.html#things-that-don-t-work).


junkmd:
I am investigating and re-running the failed jobs in the CI pipeline.

Even before this PR, memory leaks would occasionally occur, but re-running the jobs would result in success, so I had assumed these were one-time glitches.
However, with the changes in this PR, the leaks have become highly reproducible.

This project employs various kinds of metaprogramming and is also related to threading.

It will likely require thorough investigation.

moi15moi:
Why do you keep running the CI? There have been 111 attempts, and they all failed.
It's pretty clear that there is a problem. ^^'

junkmd:
In between my main work and other investigations, I was repeatedly re-running the failed jobs to see if they would pass.

As you pointed out, there are clearly problems.

One key observation is that while failures occur depending on whether third-party dependencies are included or based on the system’s bitness, at least one test run for each Python version has passed. This means that every configuration has the potential to pass — but also the potential to fail.

We should continue investigating the root cause, but to prevent memory leaks and stabilize the tests, we may need to escalate this to `CPython`, or open an issue in `coverage-py`.

junkmd:
The coverage measurement of the test codebase might be affecting the results.
Try setting `omit = ["comtypes/gen/*", "comtypes/test/*"]` to measure coverage only for the production codebase.

moi15moi:
In the [documentation](https://coverage.readthedocs.io/en/7.6.10/), they recommend to include the coverage of the test.
image

junkmd:
The previous suggestion was meant to verify whether test code coverage measurement affects memory release.
The primary intent is to experiment with different conditions to investigate the root cause.

I agree that test code should be included in the coverage measurement. However, we first need to determine what is actually affecting the results.

Possibly, it’s the coverage measurement itself.
Possibly, the memory leak test for the COM server is a false positive.

junkmd:
Thank you for the verification commit.

Since it fails no matter how many times I re-run it, it seems that test code coverage measurement is not the root cause of this issue.

Since the `eval` method calls Python’s `eval` function, it is possible that this test double implementation itself is inherently prone to memory leaks.

https://github.com/enthought/comtypes/blob/f1e79a9cc625aae287f14ab9db859cc048cc8204/comtypes/test/TestComServer.py#L104-L108

junkmd:
Generating a tuple might be the issue.

https://github.com/enthought/comtypes/blob/f1e79a9cc625aae287f14ab9db859cc048cc8204/comtypes/test/test_comserver.py#L93-L100

How about trying "1 + 2" instead?

junkmd:
I confirmed that changing from `"(1, 2, 3)"` to `"1 + 2"` still resulted in some job failures on Python 3.12.

[The failing test in a job](https://github.com/enthought/comtypes/actions/runs/13065140715/job/36456092110#step:6:421) was the same as before: `test_comserver.TestInproc_win32com.test_eval`.

However, what concerns me even more is that [a job where all tests passed on Python 3.12](https://github.com/enthought/comtypes/actions/runs/13065140715/job/36456092610#step:6:371) takes **818.020s**.

In contrast, [a job where all tests passed on Python 3.11](https://github.com/enthought/comtypes/actions/runs/13065140715/job/36456089636#step:6:407) only took **83.091s**, which means the execution time has increased nearly tenfold.

Looking at [a canceled job on Python 3.12](https://github.com/enthought/comtypes/actions/runs/13065140715/job/36456091024#step:6:130), I noticed that `test_client.Test_GetModule.test_the_name_of_the_enum_member_and_the_coclass_are_duplicated` was canceled.
This test generates the Python wrapper module for `MSHTML` (`_3050F1C5_98B5_11CF_BB82_00AA00BDCE0B_0_4_0.py`).

As mentioned in https://github.com/enthought/comtypes/issues/524#issuecomment-2047404531, the wrapper module for `MSHTML` is massive.
Could it be that despite excluding it from coverage measurement with `omit = ["comtypes/gen/*"]` it is still attempting to measure coverage and that’s why the execution time is so long?
It's puzzling that this issue does not occur with Python versions other than **3.12**.

Currently, the latest version (**7.6.10**) of `coverage` is installed on Python 3.12 jobs.
Perhaps we should try explicitly specifying another version.

junkmd:
These issues might be related:
https://github.com/python/cpython/issues/127953
https://github.com/python/cpython/issues/107674

junkmd:
Thank you for pinging me about this PR.

Since https://github.com/python/cpython/pull/107841 has been merged, the slowdown issue with coverage might be resolved in the latest Python 3.12.10.
However, it seems the default Python 3.12 version in GHA is still 3.12.9.

Once the default Python revision used in GHA is updated, I believe this PR should be able to be merged without any issues.

moi15moi:
I still need to implement the codecov step to fix #772
I just rebased to see if the issue with the leak was fixed.

junkmd:
@moi15moi
Thank you for your continued contributions.
My concern is the handling of `secrets.CODECOV_TOKEN`.
My understanding is that for public repositories, no special token is required.
I'm not sure if the coverage can be uploaded to Codecov because this token exists and you're opening a pull request from your fork, or if it's because you're using Enthought's token.
If there are any similar precedents or documentation regarding this kind of PR, please let me know.

@dpinte
Please share your thoughts as an Enthought member.
Is it okay to use `secrets.CODECOV_TOKEN` in the comtypes repository?
Or is this token currently not set, and the coverage is being uploaded simply due to the combination of Codecov and a public repository?

junkmd:
I understand that with the current workflow file, even without secrets defined for the Codecov token, uploading coverage from this public repository isn't a problem.
Since the scope of this PR is to enable coverage measurement, I'd like to merge this.

junkmd:
@dpinte 

It seems a token is required to upload coverage for protected branches like `main`.

```
error - 2025-06-03 08:07:29,281 -- Upload failed: {"message":"Token required because branch is protected"}
```
https://github.com/enthought/comtypes/actions/runs/15411905408/job/43365628109#step:7:699

Please set up the token.

dpinte:
@junkmd a CODECOV_TOKEN secret has been added to the repository. You can use it now for uploading reports

junkmd:
@dpinte 

Thanks for adding the secret to this repository. I re-ran the job and confirmed that the coverage was successfully uploaded.

On a related note, I noticed that the badge in the README, [Coverage](https://app.codecov.io/github/enthought/comtypes), doesn't display the coverage percentage correctly. I plan to swap it out for [Coverage](https://app.codecov.io/github/enthought/comtypes/tree/main) when I address #840.

PR #777, opened 2025-01-28
[_post_coinit\unknwn.py] Remove unused parameter in _shutdown

moi15moi:
sys.exc_clear() was removed in Python 3.0. For more details, refer to the Python 3.0 release notes: https://docs.python.org/3/whatsnew/3.0.html

### DIFF
--- a/comtypes/_post_coinit/unknwn.py
+++ b/comtypes/_post_coinit/unknwn.py
@@ -1,7 +1,6 @@
 # https://learn.microsoft.com/en-us/windows/win32/api/unknwn/
 
 import logging
-import sys
 from ctypes import HRESULT, POINTER, byref, c_ulong, c_void_p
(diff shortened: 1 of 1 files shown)


junkmd (approved):
Indeed, [`sys.exc_clear`](https://docs.python.org/2.7/library/sys.html#sys.exc_clear)  no longer exists in Python 3 (source: The _Cleanup of the sys module_ part of
[_Library Changes_ section](https://docs.python.org/3/whatsnew/3.0.html#library-changes) in _What’s New In Python 3.0_).

Considering that `getattr(sys, "exc_clear", lambda: None)` was originally written by the project’s originator, it is now dead code in the Python 3-only era.

Thank you for your observation and investigation.

junkmd:
And LUCKEY 777 😄 🎉 !

You did it!

PR #778, opened 2025-01-29
Delete preview.bat

moi15moi:
comtypes requires Python 3.8 or later.
This file uses Python 2.5, and the last release that supported it was version 0.6.4, released on January 24, 2014. Therefore, it has been more than 10 years since this file was usable, so it should be safe to delete.


junkmd (approved):
Thank you for discovering the dead code.

I plan to merge it once the remaining tasks are completed.

PR #788, opened 2025-02-01
[test\test_BSTR.py] Use WinDLL/OleDLL instead of windll/oledll

moi15moi:
Note that I removed the function ``test_paramflags``.

To know why, see the commit comment.

### DIFF
--- a/comtypes/test/test_BSTR.py
+++ b/comtypes/test/test_BSTR.py
@@ -1,5 +1,6 @@
 import unittest
-from ctypes import WINFUNCTYPE, c_uint, c_void_p, oledll, windll
+from ctypes import WinDLL
+from ctypes.wintypes import UINT
 
 from comtypes import BSTR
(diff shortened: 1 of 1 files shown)


junkmd (approved):
Thanks!

I understand from your commit message why you are proposing to remove `test_paramflags` from this project.

Indeed, it essentially tests the same thing as `test_inargs`.

The functionality of `prototype`, as well as `SysStringLen` from `oledll.oleaut32` (`OleDLL("oleaut32")`) and `BSTR`, should probably be tested on the CPython side.

CPython already has [tests for `BSTR`](https://github.com/python/cpython/blob/cf4c4ecc26c7e3b89f2e56893260a8a3319dab3d/Lib/test/test_ctypes/test_bytes.py#L58-L63), and since it is important for Windows, I am considering adding more `BSTR` tests to CPython.

Issue #118, opened 2017-03-14
If `PATH` is not defined, importing `comtypes` fails on Windows


maxbelanger:
We've noticed that on some (rare) Windows devices, if `PATH` is not set or not readable when the application launches, `comtypes` crashes on import. The following stack trace illustrates the issue:

```
Traceback (most recent call last):
  <...>
  File "comtypes\client\__init__.pyc", line 21, in <module>
  File "comtypes\client\_events.pyc", line 8, in <module>
  File "comtypes\client\_generate.pyc", line 18, in <module>
KeyError: 'PATH'
```

Instead of [expecting `PATH` to be defined](https://github.com/enthought/comtypes/blob/master/comtypes/client/_generate.py#L18), the code in `_generate.py` should fall back gracefully rather than break imports.

PR #836, opened 2025-04-22
[client\_generate.py] Remove PATH variable

moi15moi:
Fix: https://github.com/enthought/comtypes/issues/118

The variable isn't used since 2008: https://github.com/enthought/comtypes/commit/a6bee92cf1ff0e3c8eec07259bf3cf22df9f5398#diff-43f06e41bd1e46dbb0ded91a220ae60e7961e2b72126ad3b1031065c0bad4fb6

### DIFF
--- a/comtypes/client/_generate.py
+++ b/comtypes/client/_generate.py
@@ -15,8 +15,6 @@ from comtypes.tools import codegenerator, tlbparser
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


junkmd (approved):
Thank you!

I've confirmed that this `PATH` constant is definitely dead code.

Since `_generate.py` has a name starting with an underscore, it's highly unlikely that users are referencing it externally.
Even if they were, the impact would be minimal, as it would only require adding a single line to that developer's codebase.

Issue #655, opened 2024-11-04
Remove the dead code that performs an early return with `if bases == (object,)` in `_coclass_meta`.


junkmd:
Whether it’s a remnant of Python 2.x or a legacy from the early stages of this project, there is dead code as shown below.

https://github.com/enthought/comtypes/blob/d80810b50eb642053b8dc1c3a2a4735bedf0a059/comtypes/_meta.py#L44-L62

If it can be confirmed that removing it poses no problems, I believe the codebase would be more readable without this section.

```diff
     # the coclass' _com_interfaces_ list.
     def __new__(cls, name, bases, namespace):
         self = type.__new__(cls, name, bases, namespace)
-        if bases == (object,):
-            # HACK: Could this conditional branch be removed since it is never reached? 
-            # Since definition is `class CoClass(COMObject, metaclass=_coclass_meta)`,  
-            # the `bases` parameter passed to the `_coclass_meta.__new__` would be      
-            # `(COMObject,)`.
-            # Moreover, since the `COMObject` derives from `object` and does not specify
-            # a metaclass, `(object,)` will not be passed as the `bases` parameter      
-            # to the `_coclass_meta.__new__`.
-            # The reason for this implementation might be a remnant of the differences  
-            # in how metaclasses work between Python 3.x and Python 2.x.
-            # If there are no problems with the versions of Python that `comtypes`
-            # supports, this removal could make the process flow easier to understand.
-            return self
         # XXX We should insist that a _reg_clsid_ is present.
         if "_reg_clsid_" in namespace:
             clsid = namespace["_reg_clsid_"]
```

PR #846, opened 2025-06-11
Run ruff check UP004

moi15moi:
I simply runned the command: ruff check --select UP004 --fix 
For more detail about UP004, see https://docs.astral.sh/ruff/rules/#pyupgrade-up
ruff version used: 0.11.13

### DIFF
--- a/comtypes/_comobject.py
+++ b/comtypes/_comobject.py
@@ -144,7 +144,7 @@ _PostQuitMessage.argtypes = [INT]
 _PostQuitMessage.restype = None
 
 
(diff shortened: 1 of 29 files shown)


junkmd (approved):
This PR can be merged as is.

Removing `object` as a base class from `COMObject` doesn't break any tests, and the tests continue to pass without issues.
As previously discussed in #655, it appears that a part of `_coclass_meta` might indeed be dead code.
If you have the time, I would appreciate it if you could investigate this further.

PR #847, opened 2025-06-12
Run ruff check UP008

moi15moi:
I simply runned the command: ruff check --select UP008 --fix --unsafe-fixes
For more detail about UP008, see https://docs.astral.sh/ruff/rules/super-call-with-parameters/
ruff version used: 0.11.13

Note that this is marked as a "unsafe-fixes" only because it can remove comments (which hasn't been the case).

### DIFF
--- a/comtypes/_comobject.py
+++ b/comtypes/_comobject.py
@@ -218,7 +218,7 @@ class COMObject:
     _dispimpl_: Dict[Tuple[comtypes.dispid, int], Callable[..., Any]]
 
     def __new__(cls, *args: Any, **kw: Any) -> "hints.Self":
-        self = super(COMObject, cls).__new__(cls)
+        self = super().__new__(cls)
         if isinstance(self, c_void_p):
             # We build the VTables only for direct instances of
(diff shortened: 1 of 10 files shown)


junkmd:
Several years ago, we discussed this at https://github.com/pywinauto/pywinauto/pull/1232, and I'm still not keen on adding CI settings that use `pywinauto` to `comtypes`.

I believe that monitoring whether a project breaks due to dependency changes should be handled within that project itself.
For example, `comtypes` depends on `ctypes`, but core developers would likely not allow adding tests to CPython that install `comtypes`.
My thinking is equivalent to that.

Currently, I'm manually triggering pywinauto's CI largely due to personal interest.

If you know of a way for `pywinauto` to detect updates to `comtypes` or other dependencies and run CI using a bot or GitHub Actions, I'd appreciate it if you could join the discussion happening within the `pywinauto` project.

junkmd (approved):
Thanks.

I also checked the patch coverage on codecov, and I don't see any particular problems given the changes.

Issue #480, opened 2023-05-18
Modernising the dev tooling


bennyrowland:
Hello `comtypes`ers,

As `comtypes` is currently experiencing a new lease of life (thanks particularly to the efforts of @junkmd), I wanted to gauge the level of interest from contributors in adding in new tooling to support the development process. There is now such a rich ecosystem of packaging tools including things like `tox`/`nox` for isolating and automating test builds, formatters like `black` (that can be run with `pre-commit`), linters like `pylint` or `ruff`, `setuptools_scm` for version management, and `towncrier` for generating the changelog.

While I am not suggesting the immediate or wholesale adoption of all these options, I think that incorporating some of these tools could make the development process easier for new contributors to get started with, and probably simplify things for everybody. However, I realise that this would also require a certain amount of work, and also some discussion to try and achieve consensus on what tools to use, and that may not appeal to everyone involved. So I decided to open this issue to see whether I get enthusiastic support, naked hostility, or studied indifference. Let me know what you all think...

junkmd:
Thank you so much for the compliment, @bennyrowland!

This issue is worth to be pinned (so I actually did).

I will summarize my opinions during my free time this weekend.

junkmd:
First of all, most of the modern tools you have exemplified are for Python3 only.

So if we were to adopt some of them, it would be the `dorp_py2` branch, not the `master` branch which still supports Python 2.7.

junkmd:
## `black`
This has already been partially introduced in the `drop_py2` branch(#418).
[The GHA checks if the PRed code is in the `black==22.12.0` format](https://github.com/enthought/comtypes/blob/532c3996e73866009249e02981d3ead9ae58d474/.github/workflows/autofmt.yml#L24-L25).

I am contributing to [`typeshed`](https://github.com/python/typeshed/).
In that repository, GHA runs auto-format every PR using `black`.
I thought that workflow to be excellent.

I tried to introduce it in `comtypes`.
But it was difficult for bots to add commits with `git-auto-commit` and `pre-commit ci` due to permissions.
So we are currently limiting it to style checks only.

If other (perhaps at the `Member` level) maintainers agree to give these bots/actions commit privileges, I think we can introduce auto-formatting as well.

bennyrowland:
[...]

PR #848, opened 2025-06-12
Replace setup.py with pyproject.toml

moi15moi:
Fix https://github.com/enthought/comtypes/issues/480#issuecomment-1565784242
Note that when building the dist (with ``python -m build --sdist``), it won't create a zip file anymore. It will be a tar.gz file.
There is an issue about this, see https://github.com/pypa/setuptools/issues/3916, but honestly, I don't think it really matters.


moi15moi:
I'm currently trying to investigate why the install-tests CI fails only for Python 3.8

moi15moi:
Oh, the problem is simple.
The ``license`` in the pryproject.toml have been added in setuptools [77.0.0](https://setuptools.pypa.io/en/latest/history.html#v77-0-0).
The last version to support Python 3.8 is [75.3.2](https://pypi.org/project/setuptools/75.3.2/), so it cannot simply work.

I guess I will need to wait that you drop 3.8 support.

junkmd:
The migration of `setup.cfg` to `pyproject.toml`, as per this PR's title, is clear and understandable.

However, I believe the legacy `setup.py` code should be addressed in a separate scope. 
I'm not deeply familiar with packaging, so please inform me if these two tasks are interdependent.

junkmd:
@moi15moi 

I apologize for the delay in my response; I've been quite busy.

Regarding `setup.py`, you're right. My initial thought to keep `setup.py` was mainly because I wasn't entirely clear on what those two `cmdclass`es were doing.
I appreciate the clarification on `post_install`. I now understand that it's no longer necessary in modern packaging, and I'm fine with its removal.

However, I still have some uncertainties about the `test` class.
My understanding is that options like `tests=` and `use-resources=` seem to have very little meaning given the current use of test resources and the unittest suite.
On the other hand, I don't fully understand the `refcounts` part.

The `refcounts` option appears to be doing something with `sys.gettotalrefcount()`, which is a function only available in Python's debug builds. However, the specific purpose of this test remains unclear to me.
If you have any insights or conjectures about the behavior or intent of this `test` class, I would greatly appreciate it if you could share them.

moi15moi:
I tried to run ``python setup.py test`` in the CI. You can see the result here: https://github.com/moi15moi/comtypes/pull/10/checks
As you can see, it currently simply fails. I guess it is was a way to detect if there is memory leak when executing test?

moi15moi:
``license-files`` has been introduced with [77.0.0](https://setuptools.pypa.io/en/latest/history.html#v77-0-0), so I cannot directly use a older version.
It is now recommand to use ``license-files``, so, in my opinion, we should just wait to drop support for python 3.8 before merging this PR.

pyproject.toml
```diff
@@ -1,7 +1,63 @@
+[project]
+name = "comtypes"
+description = "Pure Python COM package"
+readme = "README.md"
+requires-python = ">=3.8"
+license = "MIT"
```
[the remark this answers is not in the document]
junkmd: The clauses in `LICENSE.txt` are the same as those of [the MIT License](https://opensource.org/license/mit).

Although starting the file with the line `MIT License` is the practice generally accepted today, the wording might be consistent, as the MIT License is officially listed as an OSI-approved open source license.

https://github.com/enthought/comtypes/blob/8d407153a929ea099c02e2adac2570d85b9b1e76/LICENSE.txt#L1-L2

@cfarrow
If you remember, I would like to confirm the intent behind this license wording with you, who committed `LICENSE.txt` in 8d40715.
Was this common practice when this commit was made?

pyproject.toml
```diff
@@ -1,7 +1,63 @@
...
+    { name = "Thomas Heller", email = "theller@python.net" },
+]
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Operating System :: Microsoft :: Windows",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
```
junkmd: We might migrate the `License :: OSI Approved :: MIT License` setting to `pyproject.toml`'s `classifiers`, as it was previously configured in `setup.py` or `setup.cfg`.

https://github.com/enthought/comtypes/blob/938fbd50d2077c6bd524790a9ed798c3990cdfc0/setup.py#L18-L27

https://github.com/enthought/comtypes/blob/fb8fd5823d83f866df43ce6f15d1b8dfd7f51f8c/setup.cfg#L11-L18

.github/workflows/autotest.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           architecture: ${{ matrix.architecture }}
       - name: install comtypes
         run: |
-          pip install --upgrade setuptools
+          pip install --upgrade setuptools build
```
Avasam: With PEP517 build isolation and `setuptools` specified in your `[build-system].requires` pre-installing setuptools here should do nothing.
```suggestion
          pip install --upgrade build
```

pyproject.toml
moi15moi: If i do that and run ``python -m build``, I get this error:
```
setuptools.errors.InvalidConfigError: License classifiers have been superseded by license expressions (see https://peps.python.org/pep-0639/). Please remove:

License :: OSI Approved :: MIT License

ERROR Backend subprocess exited when trying to invoke get_requires_for_build_sdist
```

This is why I didn't included it.

PR #850, opened 2025-06-13
Run ruff check UP012

moi15moi:
I simply runned the command: ruff check --select UP012 --fix
For more detail about UP012, see: https://docs.astral.sh/ruff/rules/unnecessary-encode-utf8/
ruff version used: 0.11.13


junkmd (approved):
This is easy to review and merge.

PR #851, opened 2025-06-13
Run ruff check UP024

moi15moi:
I simply runned the command: ruff check --select UP024 --fix
For more detail about UP024, see: https://docs.astral.sh/ruff/rules/os-error-alias/
ruff version used: 0.11.13

### DIFF
--- a/comtypes/_comobject.py
+++ b/comtypes/_comobject.py
@@ -414,7 +414,7 @@ class COMObject:
         try:
             self.__typelib
         except AttributeError:
-            raise WindowsError(hresult.E_NOTIMPL)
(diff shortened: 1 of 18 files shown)


junkmd (approved):
I've confirmed [`EnvironmentError`](https://docs.python.org/3/library/exceptions.html#EnvironmentError), [`IOError`](https://docs.python.org/3/library/exceptions.html#IOError), and [`WindowsError`](https://docs.python.org/3/library/exceptions.html#WindowsError) have been replaced by [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError).

While aliases might still exist if searched manually, I approve as applying `UP024` is the scope of this PR.

Issue #643, opened 2024-10-21
Reducing `tool.ruff.lint.per-file-ignores`.


junkmd:
We are looking to reduce the number of rules ignored in `tool.ruff.lint.per-file-ignores` by delinting the codebase. This is a great opportunity for new contributors to get involved!

**Guidelines:**
- We welcome delinting of one rule per file, or keeping changes as small as possible per pull request, as it makes it easier for reviewers to check the changes.
- If your PR addresses multiple rules or files, reviewers might consider it a large change and suggest splitting it into multiple PRs.
- When modifying runtime codebases, be careful not to break backward compatibility. For example, when delinting `E722` 'bare-except', specifying the type of exception to catch may result in different behavior before and after the change.
- If you want to delint without changing the implementation, use the minimal `# noqa: RULE` comments.

junkmd:
With the codebase changes up to commit 01ec70d, the following should definitely be achievable.
I’ll leave these changes to new contributors.

```diff
 [tool.ruff.lint.per-file-ignores]
 # production
-"comtypes/_comobject.py" = ["E713", "E722", "F401"]
 "comtypes/_npsupport.py" = ["F401"]
 "comtypes/_vtbl.py" = ["E722"]
 "comtypes/connectionpoints.py" = ["F401", "F403", "F405"]
```

```diff
 "comtypes/shelllink.py" = ["F401"]
 "comtypes/util.py" = ["F403", "F405"]
 "comtypes/viewobject.py" = ["F403", "F405"]
-"comtypes/_post_coinit/unknwn.py" = ["F821"]
 "comtypes/client/__init__.py" = ["F401", "F403"]
 "comtypes/client/_code_cache.py" = ["E401", "E711"]
 "comtypes/client/_constants.py" = ["F401"]
```

```diff
 "comtypes/server/__init__.py" = ["E401"]
 "comtypes/server/automation.py" = ["F403", "F405"]
 "comtypes/server/connectionpoints.py" = ["F401", "F403", "F405"]
-"comtypes/server/inprocserver.py" = ["E713", "E722", "F403", "F405", "F841"]
+"comtypes/server/inprocserver.py" = ["E713", "E722", "F841"]
 "comtypes/server/localserver.py" = ["F401", "F403", "F405"]
-"comtypes/server/register.py" = ["F403", "F405", "E713", "E731"]
+"comtypes/server/register.py" = ["F403", "F405", "E713"]
[...]

PR #943, opened 2026-07-02
Remove F401 (unused import) ignores, fix all violations

Mubashir78:
Part of #643

- Remove 12 unused imports found by ruff F401 across the codebase
- Remove F401 from all 13 `per-file-ignores` entries in pyproject.toml
- Drop per-file-ignores lines that had only F401


comtypes/_npsupport.py
```diff
@@ -157,7 +157,6 @@ def enable(self):
         if self.enabled:
             return
         # first we have to be able to import numpy
-        import numpy
```
junkmd, line 160: As noted in the comment above, we need to import `numpy` here to cache it in the Python import system.

By adding [`# noqa: F401`](https://docs.astral.sh/ruff/rules/unused-import/) only to this line, you can remove the setting from `pyproject.toml` while keeping the import.

docs/source/conf.py
```diff
@@ -13,7 +13,6 @@
 import os
 import sys
 
-import sphinx_rtd_theme
```
junkmd, line 16: This might have been necessary for Sphinx.
Please check if the documentation can still be built even if this is removed.

comtypes/test/setup.py
```diff
@@ -2,6 +2,5 @@
 import glob
 from distutils.core import setup
 
-import py2exe
```
junkmd, line 5: Integration between this project and `py2exe` might already be broken, but I'd like to keep it as it is for now.
Please also add `# noqa: F401` to this line.

junkmd (changes requested):
I've returned to contributing to open source.
Sorry for the delay, but I'll review your PR now.

